PR #14, opened 2015-02-28
use modularized `lodash.assign` instead of `lodash/object/extend`

deepsweet:
https://github.com/lodash/lodash/tree/3.0.0-npm-packages/lodash.assign


sebmck:
Thanks!

Issue #5, opened 2015-02-27
no-unused-vars for jsx


gpbl:
First, thank you for babel-eslint: it's like a dream :+1: 

The following code warns for `Router is defined but never used (no-unused-vars)`:

```jsx
import Router from 'react-router';
export default (
  <Router.Route name="app" path="/" />
);
```

It seems happening when using JSX outside a `React.createClass()`. Is there a way to make eslint recognizing the use of the var?

gaearon:
+1, I also have this.

sebmck:
Most of these are easy fixes my laptop shit itself so I won't be able to
fix these for a couple of hours while I get a charger...

On Saturday, 28 February 2015, Dan Abramov <notifications@github.com> wrote:

> +1, I also have this.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/babel/babel-eslint/issues/5#issuecomment-76455272>.
>


-- 
Sebastian McKenzie

sebmck:
Fixed as of 1.0.8.

esnunes:
sorry but version 1.0.8 of what package?

cesarandreu:
I think I'm having this problem as well.

**.eslintrc**
```json
{
  "parser": "babel-eslint",
  "rules": {
    "strict": false,
    "no-console": false,
    "no-multiple-empty-lines": false
  },
  "env": {
    "mocha": true,
    "browser": true,
    "node": true
  }
}
```

**problem-example.jsx**
```jsx
var Foo = require('./Foo.jsx')
module.exports = <Foo></Foo>

```

**Sublime Text console**
```
SublimeLinter: eslint output:
<text>: line 1, col 4, Error - Foo is defined but never used (no-unused-vars)
<text>: line 2, col 18, Error - 'Foo' is not defined. (no-undef)
```

**terminal**
```sh
$ npm ls -g babel-eslint eslint
/Users/cesarandreu/.nvm/versions/io.js/v1.6.1/lib
├── babel-eslint@2.0.2
└── eslint@0.17.1
```


As part of the Sublime Text console output it also shows it's using the correct eslint binary.

frankychung:
I also had issues with this after eslint 0.17. It seems we should install `eslint-plugin-react` as explained in https://github.com/eslint/eslint/issues/2147. 
With the following:
```
$ npm -g ls eslint eslint-plugin-react babel-eslint
/usr/local/lib
├── babel-eslint@2.0.2
├── eslint@0.17.1
[...]

PR #36, opened 2015-03-07
JSX AST fixes

BenoitZugmeyer:
As `espree` now parses JSX, `babel-jslint` doesn't have to transform JSX node to normal ES nodes, they can stay the same, `eslint` will happily handle those.

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -7,11 +7,19 @@ exports.toToken = function (token) {
 
   if (type === tokTypes.name) {
     token.type = "Identifier";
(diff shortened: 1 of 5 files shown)


sebmck:
The JSX node conversions are due to #5.

BenoitZugmeyer:
Thank you for your lights. I'm writing a way to write simple non-regression tests, I'll open a separate pull request soon.

BenoitZugmeyer:
This pull request should be applied after #37 because it defines new non-regression tests.

The new last commit of this pull request (f9a84aa) fixes many scope issues caused by passing the `sourceType = "module"` option to `escope`, including issue #5 . This isn't pretty code, but it's simple and straightforward, and I didn't find a better solution to this problem (and a tried a lot of things, believe me!).

sebmck:
Thanks! I'll push out a new version shortly. eslint 0.16.0 seems to be extremely buggy, then again it's pre-1.0.0.

PR #37, opened 2015-03-07
Non regression tests

BenoitZugmeyer:
It's me again, with more tests. `babel-eslint` is built on fast moving technologies, and tries to fix many edge cases users are reporting. To help people contribute, here is a simple non-regression test suit to make sure new changes won't breaks the previously fixed issues.

I read every closed issues and wrote a test when it was relevent. I checked if the test I wrote breaks when I removed the patch fixing the issue, just to be sure.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
   },
   "homepage": "https://github.com/babel/babel-eslint",
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


sebmck:
Thanks!

PR #50, opened 2015-03-16
acorn-to-esprima: fix ExportNamedDeclaration, closes #48

### LINKED ISSUE #48: 0:0  error  Cannot read property '0' of null on multiple exports
Geting the error in the title when exporting multiple things from one file:

```
export const c = 'c';

export class Store{
}
```

UltCombo:
Ref: #48

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -85,7 +85,7 @@ var astTransformVisitor = {
         } else if (node.declaration.type === "ClassExpression") {
           node.declaration.type = "ClassDeclaration";
         }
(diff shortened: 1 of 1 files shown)


UltCombo:
Oops, this breaks way too many tests. Guess it isn't as simple as I first thought.

UltCombo:
Wait, even if I undo the changes, there are still 20 failing tests. @sebmck are all of these due to breaking changes in babel-core/esprima?

UltCombo:
Looks like this is due to a change in Esprima, the top-level AST node (root) now has a `sourceType` property.

Issue #53, opened 2015-03-18
ECMAFeature classes should throw when disabled


emmenko:
I've noticed that if I use `babel-eslint` as parser, the `ecmaFeatures.classes = false` option does nothing.

Given this configuration
```bash
$ cat .eslintrc
/* http://eslint.org/docs/rules/ */
{
  "parser": "babel-eslint",
  "ecmaFeatures": {
    "classes": false
  },
  "env": {
    "node": true
  },
  "rules": {
    "eol-last": 2,
    "quotes": [2, "single"],
    "semi": 0,
    "strict": 0
  }
}
```

and this file
```javascript
// test-class.js

class Foo {
  constructor() {
    this.foo = 'bar'
  }
}

```

When I run

```bash
$ eslint test-class.js
$ echo $?
0
```

When I remove the `parser` option it works
 
```bash
$ cat .eslintrc
/* http://eslint.org/docs/rules/ */
{
  "ecmaFeatures": {
    "classes": false
  },
  "env": {
    "node": true
  },
  "rules": {
    "eol-last": 2,
    "quotes": [2, "single"],
    "semi": 0,
    "strict": 0
  }
}

$ eslint test-class.js

test-class.js
  1:1  error  Unexpected reserved word

✖ 1 problem (1 error, 0 warnings)
```

Is this a problem with `babel-eslint` ?

sebmck:
Nope, intended behaviour. Acorn does not currently support feature-specific flags.

emmenko:
Woah, that was fast ;)

So what should I do? Do you have any suggestion or workaround?

sebmck:
Can't really do anything about it.

emmenko:
So that means that any `ecmaFeatures` option is ignored?

sebmck:
Correct.

emmenko:
Hmm ok so if want to forbid the use of `class` (or something else) I should create an `eslint-plugin` ?

sebmck:
That's the only way, yes.

emmenko:
Ok, at least I know how to proceed. Many thanks for the fastest support ever! ;)

emmenko:
For who's interested, I'm creating a plugin to forbid `class`
https://github.com/emmenko/eslint-plugin-no-class

PR #54, opened 2015-03-19
Add note that ecmaFeatures are not supported

emmenko:
I thought it would be better to add a note about that. What do you think?

Refs https://github.com/babel/babel-eslint/issues/53

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,8 @@
 **NOTE:** Please note that this is experimental and may have numerous bugs. It is however
 successfuly linting the [babel core](https://github.com/babel/babel/blob/master/.eslintrc).
 
+> Currently the `ecmaFeatures` options from `eslint` are **not** supported by the `acorn` parser and thus are ignored.
+
 ## How does it work?
 
 ESLint allows custom parsers. This is great but some of the syntax nodes that Babel supports


sebmck:
Don't really think this is necessary and it's evident in both the examples and the copytext of the README that by default you'll be able to write any Babel-compatible code with no config options by default.

emmenko:
I'm sorry but I don't think it was really clear from the README that I can't use all options from `eslint`, at least not from me as you could see. Also, which examples?

sebmck:
> babel-eslint allows you to lint ALL valid Babel code

```json
{
  "parser": "babel-eslint",
  "rules": {
    "strict": 0
  }
}
```

There's no mention of needing specific options to enable certain syntax and this will just confuse more people than it's worth.

sebmck:
People are using `babel-eslint` even though eslint now supports ES6 just because the `ecmaFeatures` option is so confusing.

emmenko:
Ok then I got this wrong from the beginning. Also I just saw in `eslint` documentation this

> Note when using a custom parser, the `ecmaFeatures` configuration property is ignored unless otherwise specified by the parser.

I don't know, something was confusing me all the time. This was the first time using this new stuff and apparently I missed the all point here. 

Sorry for the confusion and thanks for the patient to explain. Hopefully I won't bother you again ;)

emmenko:
@sebmck One more question: when using `eslint` with `babel` (and `babel-eslint` of course), which environment options would you recommend using? I see there is a `es6` option. Is that needed?

Currently I have those `env` variables active

```
"env": {
  "browser": true,
  "jasmine": true,
  "node": true
}
```

sebmck:
The `es6` option would likely include es6 built-ins such as `Promise`, `Map`, `Set` etc.

emmenko:
Ok so would you recommend activating it?

sebmck:
Just add it to your `.eslintrc`?

```json
{
  "env": {
    "es6": true
 }
}
```

emmenko:
Yeah sure, I was more asking if there would be some "complications" with it. I will enable it and see how it goes. Thanks again!

PR #76, opened 2015-04-22
Attach comments to ast using estraverse

### LINKED ISSUE #60: Support for comments
I wanted to use the rule `no-warning-comments`, however the ast doesn't have the comment node when passed to ESLint.
Is there any plan to support comments?
I wanted to try it out myself, but I don't know well enough the ast structure to add node to it. I saw that all the comments are available on `ast.comments`, but are not used in `acornToEsprima`

Cellule:
fixes #60 
Greatly inspired by https://github.com/babel/babel/blob/master/src/babel/helpers/parse.js#L32

I haven't tested yet if this works for all comments, but I tried a few real quick and it seems to work. I'll try to get this running on the full ESLint test suite to see if it works on all comments related rules.

### DIFF
--- a/index.js
+++ b/index.js
@@ -52,6 +52,51 @@ function monkeypatch() {
   };
 }
 
+exports.attachComments = function(ast, comments, tokens) {
+  estraverse.attachComments(ast, comments, tokens);
+
+  if (comments.length) {
+    var firstComment = comments[0];
+    var lastComment = comments[comments.length - 1];
+    // fixup program start
+    if (!tokens.length) {
+      // if no tokens, the program starts at the end of the last comment
+      ast.range[0] = lastComment.range[1];
+      ast.loc.start.line = lastComment.loc.end.line;
+      ast.loc.start.column = lastComment.loc.end.column;
+    } else if (firstComment.start < tokens[0].range[0]) {
+      // if there are comments before the first token, the program starts at the first token
+      var token = tokens[0];
+      ast.range[0] = token.range[0];
+      ast.loc.start.line = token.loc.start.line;
+      ast.loc.start.column = token.loc.start.column;
+
+      // estraverse do not put leading comments on first node when the comment
+      // appear before the first token
+      if (ast.body.length) {
+        var node = ast.body[0];
+        node.leadingComments = [];
+        var firstTokenStart = token.range[0];
+        var len = comments.length;
+        for(var i = 0; i < len && comments[i].start < firstTokenStart; i++ ) {
+          node.leadingComments.push(comments[i]);
+        }
+      }
+    }
+    // fixup program end
+    if (tokens.length) {
+      var lastToken = tokens[tokens.length - 1];
+      if (lastComment.end > lastToken.range[1]) {
+        // If there is a comment after the last token, the program ends at the
+        // last token and not the comment
+        ast.range[1] = lastToken.range[1];
+        ast.loc.end.line = lastToken.loc.end.line;
+        ast.loc.end.column = lastToken.loc.end.column;
+      }
+    }
+  }
+};
+
 exports.parse = function (code) {
   try {
     monkeypatch();
@@ -93,7 +138,7 @@ exports.parse = function (code) {
 
   // add comments
   ast.comments = comments;
-  estraverse.attachComments(ast, comments, tokens);
+  exports.attachComments(ast, comments, tokens);
 
   // transform esprima and acorn divergent nodes
   acornToEsprima.toAST(ast);
--- a/index.js
+++ b/index.js
@@ -6,6 +6,7 @@ var parse          = require("babel-core").parse;
 var path           = require("path");
 var t              = require("babel-core").types;
 
+var estraverse;
 var hasPatched = false;
 
 function createModule(filename) {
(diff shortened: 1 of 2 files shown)


sebmck:
Does ESLint use comments in the same format?

Cellule:
There seems to be a small difference. 
for the code
```js
var a = 2; // this is a comment
/*
this is
a multiline comment
*/
module.export = a;
```
espree attaches comments as trailing comments for the var declaration AND as leading comments for the expression statement. estraverse, on the other hand, will only attach them as leading comments for the expression statement.

Other than that, their format is pretty much the same except babel-eslint would give more information than needed.
```js
espree
[ { type: 'Line',
  value: ' this is a comment',
  range: [ 11, 31 ],
  loc: { start: { line: 1, column: 11 }, end: { line: 1, column: 31 } } },
{ type: 'Block',
  value: '\nthis is\n a multiline comment',
  range: [ 32, 65 ],
  loc: { start: { line: 2, column: 0 }, end: { line: 4, column: 22 } } } ] } ]
```

```js
babel
[ { type: 'Line',
  value: ' this is a comment',
  start: 11,
  end: 31,
  loc: { start: { line: 1, column: 11 }, end: { line: 1, column: 31 } },
  range: { '0': 11, '1': 31 },
  extendedRange: [ 10, 66 ] },
{ type: 'Block',
  value: '\nthis is\n a multiline comment',
  start: 32,
  end: 65,
  loc: { start: { line: 2, column: 0 }, end: { line: 4, column: 22 } },
  range: { '0': 32, '1': 65 },
  extendedRange: [ 10, 66 ] } ]
```

Cellule:
After some research, I found that even though espree puts the trailing and leading comments, eslint emits only one of them. https://github.com/eslint/eslint/blob/master/lib/eslint.js#L487

I am not certain, but I think it could be an issue with events `LineComment:exit` and `BlockComment:exit` which don't seem to be used at all.
https://github.com/eslint/eslint/search?utf8=%E2%9C%93&q=LineComment%3Aexit&type=Code

Cellule:
I also ran the eslint test suite with babel-eslint and with my change it passed from 138 failing test cases to 102.
The rules that are no longer failing are
- no-fallthrough
- no-inline-comments
- padded-blocks
- spaced-line-comment
- valid-jsdoc

And the remaining failing test cases are
- brace-style
- comma-spacing
- consistent-return
- default-case
- indent
- key-spacing
- no-delete-var
- no-div-regex
- no-dupe-args
- no-empty-class
- no-empty
- no-extend-native
- no-extra-parens
- no-multi-spaces
- no-multi-str
- no-octal-escape
- no-octal
- no-regex-spaces
- no-sequences
- no-shadow-restricted-names
- no-shadow
- no-undef
- no-with
- quotes
- space-after-keywords
- space-infix-ops
- wrap-regex
none of which are related to comments.

Cellule:
I encountered a problem where estraverse do not understand jsx nodes. However, I am confused as to why since at https://github.com/babel/babel/blob/master/src/babel/helpers/parse.js#L32 there is no extra processing. I think it has to do with transformers but I don't understand them very well.

sebmck:
Check patch.js and the dodgy shit that's done to add visitor keys.

On Wednesday, 22 April 2015, Michael Ferris <notifications@github.com>
wrote:

> I encountered a problem where estraverse do not understand jsx nodes.
> However, I am confused as to why since at
> https://github.com/babel/babel/blob/master/src/babel/helpers/parse.js#L32
> there is no extra processing. I think it has to do with transformers but I
> don't understand them very well.
>
> --
> Reply to this email directly or view it on GitHub
> <https://github.com/babel/babel-eslint/pull/76#issuecomment-95326600>.
>


-- 
Sebastian McKenzie

Cellule:
Thanks, that was even more simple than I thought. The code already mokeypatched estraverse. I only had to use that version instead of requiring a fresh one.

Cellule:
So this solution seems sufficient to cover most case. However, there are still some issues.
```js
function bar() {
  // comment
}
```
in this example, the comment is not added to the ast (no clue why).
The difference between the 2 ast is significant, but eslint seem to understand it nevertheless.

I think there would be a need to rewrite the `attachComment` function from estraverse completely for the 2 ast to be the same.

Cellule:
@sebmck do you think this is a good solution ? I have been using this for a time now and so far I didn't see any issues with comments.

index.js
```diff
@@ -51,6 +52,51 @@ function monkeypatch() {
...
+  if(comments.length) {
+    var firstComment = comments[0];
+    var lastComment = comments[comments.length - 1];
+    // fixup program start
+    if(!tokens.length) {
+      // if no tokens, the program starts at the end of the last comment
+      ast.range[0] = lastComment.range[1];
+      ast.loc.start.line = lastComment.loc.end.line;
+      ast.loc.start.column = lastComment.loc.end.column;
+    } else if(firstComment.start < tokens[0].range[0]) {
```
sebmck: Spaces before `(` please.

sebmck:
Extremely sorry for letting this sit for so long. Just fix the styling nits and I'll happily merge, thanks!

Cellule:
Don't worry, I understand you're a busy man :) 
I think this covers the styling issues.

sebmck:
Released as of 3.1.0, thanks!

PR #89, opened 2015-05-10
Fix flow import type

tsing:
```
import type Foo from 'foo';
```

with eslint 0.21.0, the code above will cause error: 

```
error  Foo is defined but never used  no-unused-vars
```

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -69,6 +69,9 @@ var astTransformVisitor = {
     // modules
 
     if (t.isImportDeclaration(node)) {
+      if (node.isType) {
+        return this.remove();
+      }
       delete node.isType;
     }
 
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
   },
   "homepage": "https://github.com/babel/babel-eslint",
   "devDependencies": {
-    "eslint": "^0.18.0",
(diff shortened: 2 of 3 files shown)


sebmck:
Isn't this error correct? You need to use the `Foo` type somewhere?

tsing:
@sebmck 
``
import type Foo from 'foo';

export function bar(foo: Foo) {
  console.log(foo);
}
```
In the code above, type `Foo` is used, but eslint still complain ```Foo is defined but never used```

sebmck:
Not sure, I'll have to think about it. I'd rather also lint type annotations if possible.

tsing:
@sebmck Do you have any ideas about this ?

sebmck:
Ideally flow types should be retained (ie. remove [this](https://github.com/babel/babel-eslint/blob/master/acorn-to-esprima.js#L77) line) and ESLint internals patched to crawl the type annotations and report accurately on them.

tsing:
@sebmck It seems like a big work to do,  I'm noob on this kind of AST things. 

Do you have plan to implement this ? 

Or could you implement this for now, since there is already a line remove all the type annotations?

tsing:
close this as @sebmck point out this may break peoples linting as some people have it setup with their testing frameworks etc.

PR #90, opened 2015-05-11
change more tokentypes to be punctuator - fixes #59

### LINKED ISSUE #59: `space-infix-ops` is broken
```
var i = 0, a = 1, b = 2;
console.log(i-a);
console.log(i+a);
console.log(i>a);
console.log(i<a);
console.log(i<=a);
console.log(i>=a);
console.log(i>=a);
console.log(i^a);
console.log(i?a:b);
```

This produces no warnings when using babel-eslint as a parser and the `space-infix-ops` rule.
However, the rule will still produce warnings with operators such as `*`, `=`, `+=`, etc.

hzoo:
#59.

Came across as similar issue in jscs as well. Some of the rules are checking something like `op.type === "Punctuator"` but some of the types are still `TokenType` rather than a string.

I basically just added everything in https://github.com/babel/babel/blob/master/src/acorn/src/tokentype.js#L48-L96 that wasn't already there.

I'm not sure we want all of these though?

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -13,7 +13,16 @@ exports.toToken = function (token) {
              type === tt.slash || type === tt.dot ||
              type === tt.bracketL || type === tt.bracketR ||
              type === tt.ellipsis || type === tt.arrow ||
-             type === tt.star ||
+             type === tt.star || type === tt.incDec ||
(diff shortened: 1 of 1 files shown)


sebmck:
Nope, definently want them all. Thanks!

Issue #31, opened 2015-03-04
comma in template string right after ${} causes comma-spacing


dimsmol:
minimal example:
```
let a = 1;
console.log(`${a}, b`); // will report comma-spacing
```
comma escaping can be used as a workaround:
```
let a = 1;
console.log(`${a}\, b`); // no error
```

gaearon:
Yep, same problem in JSX text content.

douglasduteil:
Same

mkawalec:
We have the same problem.

sebmck:
Can you please verify that this issue still exists on 3.0.1?

wavded:
Using 3.0.1 I still encounter the issue with commas either in template strings or in JSX.

abelnation:
I still see it as well.

src line:
`console.log(`${ios}, ${android}, ${phone}, ${tablet}`)`

error:
`  267:27  error  A space is required after ','  comma-spacing`

zdychacek:
Still not working, the same error:

```log.info(`Failed to enable text ad ${advertId}, trying to enable banner.`, textAdException);```

**Comment by atombender** [comment]
Same here, still getting it with 3.0.1.

**Comment by hzoo** [comment]
Looking into this... The tokens in espree/esprima aren't the same?

```js
// espree (has no comma node) and thus no issue
[ { type: 'Template', value: '`${' },
  { type: 'Identifier', value: 'a' },
  { type: 'Template', value: '}, b`' },
  { type: 'Punctuator', value: ';' } ]
// babel/acorn
[ { type: 'Punctuator', value: '`' },
  { type: 'Punctuator', value: 'template' },
  { type: 'Punctuator', value: '${' },
  { type: 'Identifier', value: 'a' },
  { type: 'Punctuator', value: '}' },
  { type: 'Punctuator', value: ', b' },
  { type: 'Punctuator', value: '`' },
  { type: 'Punctuator', value: ';' } ]
```
So maybe we need to combine tokens to make a `Template` type? @sebmck

hzoo:
Ok `comma-spacing` in jsx text should be fixed in 3.1.5.

MoOx:
I just tried 3.1.5 and still get issue on this snippet


```js
                  (
                    `upload/`,
                    `upload/` +
                      `w_${ Math.round(item.crop.width / item.crop.ratio) },` +
                      `h_${ Math.round(item.crop.height / item.crop.ratio) }/` +
[...]

PR #93, opened 2015-05-12
add tokenTypes: Null, Boolean, RegularExpression

hzoo:
Looking at the token types
in espree: https://github.com/eslint/espree/blob/master/lib/token-info.js#L57-L68
in esprima: https://github.com/jquery/esprima/blob/master/esprima.js#L85-L94.

Only additions are `JSXText` and `JSXIdentifier` in espree.

This adds `Null`, `Boolean`, `RegularExpression`.

Not sure about the `Template` token type since it seems that we would have to combine tokens? https://github.com/babel/babel-eslint/issues/31#issuecomment-100912106

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -34,6 +34,10 @@ exports.toToken = function (token) {
     token.value = ">";
   } else if (type === tt.jsxName) {
     token.type = "JSXIdentifier";
+  } else if (type.keyword === "null") {
+    token.type = "Null";
+  } else if (type.keyword === "false" || token.keyword === "true") {
+    token.type = "Boolean";
   } else if (type.keyword) {
     token.type = "Keyword";
   } else if (type === tt.num) {
@@ -42,6 +46,8 @@ exports.toToken = function (token) {
   } else if (type === tt.string) {
     token.type = "String";
     token.value = JSON.stringify(token.value);
+  } else if (type === tt.regexp) {
+    token.type = "RegularExpression";
   }
 
   return token;


sebmck:
Thanks!

> Not sure about the `Template` token type since it seems that we would have to combine tokens? 

Most likely, yeah.

hzoo:
So fast! So that would have to be a transformation step after `parse(code, opts)` and before the `ast.tokens = tokens.map(acornToEsprima.toToken);`?

sebmck:
It'd be easier to add an `acornToEsprima.toTokens` method and replace the `map(acornToEsprima.toToken)`.

hzoo:
Could I get some help on how to do this?

Right now my naive idea would just be searching the `tokens` array for ``token.value === '`' ``, then checking if the next ``token.value === 'template'``, and then similarly for ``token.value: '${'`` - and if so combine into a single token. And similar idea for the ending.

sebmck:
Don't have time to dig into this right now, sorry.

Here are the relevant parsing functions from Acorn, you can see how it "picks" tokens and chooses when to end a template literal:

```javascript
pp.parseTemplateElement = function() {
  let elem = this.startNode()
  elem.value = {
    raw: this.input.slice(this.start, this.end),
    cooked: this.value
  }
  this.next()
  elem.tail = this.type === tt.backQuote
  return this.finishNode(elem, "TemplateElement")
}

pp.parseTemplate = function() {
  let node = this.startNode()
  this.next()
  node.expressions = []
  let curElt = this.parseTemplateElement()
  node.quasis = [curElt]
  while (!curElt.tail) {
    this.expect(tt.dollarBraceL)
    node.expressions.push(this.parseExpression())
    this.expect(tt.braceR)
    node.quasis.push(curElt = this.parseTemplateElement())
  }
  this.next()
  return this.finishNode(node, "TemplateLiteral")
}
```

hzoo:
That's cool, something to help me get started is probably helpful enough

hzoo:
Ok found some eslint rules were broken because of this - going to fix and actually add tests for this.

acorn-to-esprima.js
```diff
@@ -34,6 +34,10 @@ exports.toToken = function (token) {
     token.value = ">";
   } else if (type === tt.jsxName) {
     token.type = "JSXIdentifier";
+  } else if (type.keyword === "null") {
+    token.type = "Null";
+  } else if (type.keyword === "false" || token.keyword === "true") {
```
hzoo: And wow `token.keyword === "true"` should be `type.keyword === "true"`..

PR #75, opened 2015-04-21
Ensure that escope is parsing decorators.

matteosuppo:
Address issue https://github.com/babel/babel-eslint/issues/72

PR #105, opened 2015-05-20
patch escope to visit decorators - fixes #72

### LINKED ISSUE #72: Using decorator doesn't count as usage by eslint
Have the following code:
```
import autobind from 'autobind-decorator';

class TextareaAutosize {
  ...

  @autobind
  _onChange(e) { ... }
}
```
which results into:
```
   7:7   error    autobind is defined but never used                                                no-unused-vars
```

hzoo:
Adding more to what @matteosuppo did in #75 which should fix #72?

Referenced: https://github.com/wycats/javascript-decorators
Simplified the code to patch escope.
Patch referencer: https://github.com/estools/escope/blob/master/src/referencer.js
- visitClass method: https://github.com/estools/escope/blob/master/src/referencer.js#L353
- visitProperty method: https://github.com/estools/escope/blob/master/src/referencer.js#L384

Add tests for all including params, get/set.

~~So the tests added in `non-regression.js` still pass without the changes in `index.js`. Might be something dumb.~~ Fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,6 +54,41 @@ function monkeypatch() {
     estraverse.VisitorKeys.TypeAlias = TypeAliasKeys;
     return results;
   };
+
+  // monkeypatch escope/referencer
+  var referencerLoc;
+  try {
+    var referencerLoc = Module._resolveFilename("./referencer", escopeMod);
+  } catch (err) {
+    throw new ReferenceError("couldn't resolve escope/referencer");
+  }
+  var referencer = require(referencerLoc);
+
+  // if there are decotators, then visit each
+  function visitDecorators(node) {
+    if (!node.decorators) {
+      return;
+    }
+    for (var i = 0; i < node.decorators.length; i++) {
+      if (node.decorators[i].expression) {
+        this.visit(node.decorators[i]);
+      }
+    }
+  }
+
+  // monkeypatch referencer methods to visit decorators
+  var visitClass = referencer.prototype.visitClass;
+  referencer.prototype.visitClass = function (node) {
+    // visit decorators that are in: Class Declaration
+    visitDecorators.call(this, node);
+    visitClass.call(this, node);
+  }
+  var visitProperty = referencer.prototype.visitProperty;
+  referencer.prototype.visitProperty = function (node) {
+    // visit decorators that are in: Visit Property / MethodDefinition
+    visitDecorators.call(this, node);
+    visitProperty.call(this, node);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


hzoo:
Dang `Cannot find module './referencer'`...

So it's downloading `node_modules/escope/escope.js` rather than the actual `node_modules/escope/lib/`

index.js
```diff
@@ -54,6 +54,40 @@ function monkeypatch() {
...
+  try {
+    var referencerLoc = Module._resolveFilename("./referencer", escopeMod);
+  } catch (err) {
+    throw new ReferenceError("couldn't resolve escope/referencer");
+  }
+  var referencer = require(referencerLoc);
+
+  // if there are decotators, then visit each
+  function visitDecorators(node) {
+    if (node.decorators) {
```
druska: Could make this opposite and return early to reduce overall indentation.

hzoo: Sure sounds good. Actually wanted to make a jscs rule for this https://github.com/jscs-dev/node-jscs/issues/1223 - not sure how to implement yet.

hzoo:
Ok finally passed. Can you take a look @sebmck?

index.js
```diff
@@ -54,6 +54,41 @@ function monkeypatch() {
...
+    throw new ReferenceError("couldn't resolve escope/referencer");
+  }
+  var referencer = require(referencerLoc);
+
+  // if there are decotators, then visit each
+  function visitDecorators(node) {
+    if (!node.decorators) {
+      return;
+    }
+    for (var i in node.decorators) {
```
sebmck: Would be better to loop over this with `for (var i = 0; i < node.decorators.length; i++)` only because I've had issues in the past where people monkeypatch `Array.prototype` and add enumerable properties...

sebmck:
Other than the loop thing, looks good!

index.js
hzoo: Haha I guess since we're actually monkeypatching it's a good assumption to make!

test/non-regression.js
```diff
@@ -163,4 +163,96 @@ describe("verify", function () {
       []
     );
   });
+
+  describe("decorators #72", function () {
+    it("class declaration", function () {
+      verifyAndAssertMessages(
+        [
+          "import classDeclaration from 'decorator';",
```
hzoo: Would be nice to use template strings here - maybe should use babel

sebmck:
Awesome work @hzoo, thanks!

druska:
Huge fix. Good job!

Issue #102, opened 2015-05-16
Flow types break ESLint


Mark-Simulacrum:
`for (let a of (a: Array)) {}` is enough to break ESLint's parsing with
```
node_modules/eslint/node_modules/escope/lib/scope.js:129
    scopes = scopeManager.__nodeToScope.get(scope.block);
                                        ^
TypeError: Invalid value used as weak map key
    at WeakMap.get (native)
    at registerScope (node_modules/eslint/node_modules/escope/lib/scope.js:129:41)
    at TDZScope.Scope (node_modules/eslint/node_modules/escope/lib/scope.js:245:9)
    at new TDZScope (node_modules/eslint/node_modules/escope/lib/scope.js:585:78)
    at ScopeManager.__nestTDZScope (node_modules/eslint/node_modules/escope/lib/scope-manager.js:254:41)
    at Referencer.materializeTDZScope (node_modules/eslint/node_modules/escope/lib/referencer.js:223:35)
    at Referencer.visitForIn (node_modules/eslint/node_modules/escope/lib/referencer.js:337:26)
    at Referencer.ForOfStatement (node_modules/eslint/node_modules/escope/lib/referencer.js:595:22)
    at Referencer.Visitor.visit (node_modules/eslint/node_modules/escope/node_modules/esrecurse/esrecurse.js:109:34)
    at Referencer.Visitor.visitChildren (node_modules/eslint/node_modules/escope/node_modules/esrecurse/esrecurse.js:88:38)
```

I believe that this is because the flow type is not stripped from the input, but I am not sure. This might be an issue with `escope`, because https://github.com/estools/escope/blob/master/src/referencer.js#L170 calls `__nestTDZScope` with an incorrect function signature (two arguments instead of one). It appears that the second argument (`iterationNode`) should be passed in to `__nestTDZScope`, but I have not tested this extensively.

The current workaround is to execute ESLint like this: `babel --whitelist flow t.js | eslint --stdin`, but this is difficult to do when integrating with in-editor plugins.

hzoo:
I don't know that much about flow at the moment but I'l look into it.

In `acorn-to-esprima.js`, it seems like we remove the flow node: https://github.com/babel/babel-eslint/blob/5f8cfd35be7b02c7e505363e5ed13bac13f51028/acorn-to-esprima.js#L174-L176

The issues before related to flow were: #66, #89.
Also there's been a few requests related to escope (decorators, array comprehensions).

sebmck:
[...]

PR #109, opened 2015-05-21
visit flow types - fixes #108

### LINKED ISSUE #108: Flow `import type` counted as "defined but never used"
```
import type Foo from 'foo';

function log(foo: Foo) {
  console.log(foo);
}
```
Complains that `Foo` is "defined but never used"  (`no-unused-vars`).

PR #89 was previously submitted but withdrawn to address this, but the problem's still there. Filing this issue to make sure we don't lose track of it.

hzoo:
#108

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -187,8 +187,22 @@ var astTransformVisitor = {
       return node.argument;
     }
 
-    if (this.isTypeCastExpression()) {
-      return node.expression;
+    // prevent "no-undef"
+    // for "Component" in: "let x: React.Component"
+    if (this.isQualifiedTypeIdentifier()) {
+      delete node.id;
+    }
+    // for "b" in: "var a: { b: Foo }"
+    if (this.isObjectTypeProperty()) {
+      delete node.key;
+    }
+    // for "indexer" in: "var a: {[indexer: string]: number}"
+    if (this.isObjectTypeIndexer()) {
+      delete node.id;
+    }
+    // for "param" in: "var a: { func(param: Foo): Bar };"
+    if (this.isFunctionTypeParam()) {
+      delete node.name;
     }
 
     // flow
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -117,7 +117,7 @@ function convertTemplateType(tokens) {
   // create Template token
   function replaceWithTemplateType(start, end) {
     var templateToken = {
-      type: 'Template',
+      type: "Template",
       value: createTemplateValue(start, end),
       range: [tokens[start].start, tokens[end].end],
       loc: {
@@ -183,16 +183,21 @@ var astTransformVisitor = {
       delete node.argument;
     }
 
+    if (this.isRestElement()) {
+      return node.argument;
+    }
+
     if (this.isTypeCastExpression()) {
       return node.expression;
     }
 
-    if (this.isFlow()) {
-      return this.remove();
-    }
+    // flow
 
-    if (this.isRestElement()) {
-      return node.argument;
+    if (this.isDeclareModule() ||
+        this.isDeclareClass() ||
+        this.isDeclareFunction() ||
+        this.isDeclareVariable()) {
+      return this.remove();
     }
 
     // modules
--- a/index.js
+++ b/index.js
@@ -76,18 +76,89 @@ function monkeypatch() {
     }
   }
 
-  // monkeypatch referencer methods to visit decorators
+  // visit decorators that are in: ClassDeclaration / ClassExpression
   var visitClass = referencer.prototype.visitClass;
   referencer.prototype.visitClass = function (node) {
-    // visit decorators that are in: Class Declaration
     visitDecorators.call(this, node);
     visitClass.call(this, node);
   }
+  // visit decorators that are in: Property / MethodDefinition
   var visitProperty = referencer.prototype.visitProperty;
   referencer.prototype.visitProperty = function (node) {
-    // visit decorators that are in: Visit Property / MethodDefinition
     visitDecorators.call(this, node);
     visitProperty.call(this, node);
+  }
+
+  function visitTypeAnnotation(node) {
+    // visit function type parameters
+    if (node.typeParameters) {
+      node.typeParameters.params.forEach(function (p) {
+        if (p.type === 'Identifier') {
+          this.visit(p);
+        } else {
+          visitTypeAnnotation.call(this, p);
+        }
+      }.bind(this));
+    }
+
+    // visit function parameters
+    if (node.params) {
+      for (var i = 0; i < node.params.length; i++) {
+        visitTypeAnnotation.call(this, node.params[i]);
+      }
+    }
+
+    // visit return type
+    if (node.returnType) {
+      visitTypeAnnotation.call(this, node.returnType);
+    }
+
+    // visit type
+    if (node.typeAnnotation) {
+      visitTypeAnnotation.call(this, node.typeAnnotation);
+    }
+
+    // visit rest
+    if (t.isFunctionTypeAnnotation(node) && node.rest) {
+      visitTypeAnnotation.call(this, node.rest);
+    }
+
+    // object properties - properties
+    if (t.isObjectTypeAnnotation(node) && node.properties) {
+      for (var i = 0; i < node.properties.length; i++) {
+        visitTypeAnnotation.call(this, node.properties[i].value);
+      }
+    }
+
+    // object properties - indexers
+    if (t.isObjectTypeAnnotation(node) && node.indexers) {
+      for (var i = 0; i < node.indexers.length; i++) {
+        visitTypeAnnotation.call(this, node.indexers[i].key);
+        visitTypeAnnotation.call(this, node.indexers[i].value);
+      }
+    }
+
+    if (node.id) {
+      this.visit(node.id);
+    }
+  }
+
+  // visit flow type in FunctionDeclaration, FunctionExpression, ArrowFunctionExpression
+  var visitFunction = referencer.prototype.visitFunction;
+  referencer.prototype.visitFunction = function (node) {
+    visitTypeAnnotation.call(this, node);
+    visitFunction.call(this, node);
+  }
+
+  // visit flow type in VariableDeclaration
+  var variableDeclaration = referencer.prototype.VariableDeclaration;
+  referencer.prototype.VariableDeclaration = function (node) {
+    if (node.declarations) {
+      node.declarations.forEach(function (d) {
+        visitTypeAnnotation.call(this, d.id);
+      }.bind(this));
+    }
+    variableDeclaration.call(this, node);
   }
 }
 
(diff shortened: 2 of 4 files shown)


acorn-to-esprima.js
```diff
@@ -187,10 +187,6 @@ var astTransformVisitor = {
       return node.expression;
     }
 
-    if (t.isFlow(node)) {
```
hzoo: So then we need to remove this line? https://github.com/babel/babel-eslint/pull/89#issuecomment-101579831

sebmck:
This *should* do the trick. At least for very simple cases, although you can have deeply nested type annotations that this wont cover.

hzoo:
So then we would need to do the same for say variable declaration type, parameter return type, etc? Yeah I guess cover the common cases?

```js
// fails at the moment
it("flow type var declaration", function () {
      verifyAndAssertMessages([
          "import type Foo from 'foo';",
          "var x: Foo = 1;",
          "x;"
        ].join("\n"),
        { "no-unused-vars": 1 },
        []
      );
    });
```

sebmck:
`escope` really is terrible for scope tracking. Yeah, those all will have to be visited too.

acorn-to-esprima.js
```diff
@@ -183,24 +183,8 @@ var astTransformVisitor = {
...
 
-    if (t.isTypeCastExpression(node)) {
-      return node.expression;
-    }
-
-    if (t.isFlow(node)) {
-      return this.remove();
-    }
-
-    if (t.isRestElement(node)) {
```
hzoo: https://github.com/babel/babel-eslint/commit/e98ee03758c60d773cbb3a33aff0af64114a270c

hzoo:
Ok this is kind of crazy - didn't cover everything but yeah.

wincent:
@hzoo: I'll ask in the internal Flow group at FB to see if anybody has any good open source examples.

wincent:
Somebody suggested the React Native example apps (eg. https://github.com/facebook/react-native/tree/master/Examples/2048 etc).

hzoo:
Ok cool I'l check it out - nice that it already has a `.eslintrc` too! (although rather old version)

hzoo:
As a result of removing

```js
if (t.isFlow(node)) {
      return this.remove();
}
```

which is needed to fix `no-unused-vars`?

---

Some more cases to fix: 
```js
// no-undef ExampleClass
type ExampleClass = {
  Component: ReactClass<any, any, any>,
  title: string,
  description: string,
};
var exampleClasses: Array<ExampleClass> = [];

// no-undef Component
var React = require('react-native');
var b = {
  openExternalExample: (null: ?React.Component)
};
var c = {
	render(): React.Component {}
};
```

---

Shouldn't be checking these..
```js
// no-undef props
var a = {
	circle: (null : ?{ setNativeProps(props: Object): void })
};

// no-undef key
var d = {
	_pressData: ({}: {[key: number]: boolean})
};

// no-undef uri, isStatic
declare module 'image!story-background' {
  declare var uri: string;
  declare var isStatic: boolean;
}
```

hzoo:
Ok rewrote most of it since there were a lot of random issues. Should of got most of the types in `Other` https://gist.github.com/hzoo/e3a7db9be7a3b0c2094e

---
Cases to fix:
```js
// have to visit TypeAlias?
// no-undef ExampleClass
type ExampleClass = {
  Component: ReactClass<any, any, any>,
  title: string,
  description: string,
};
var exampleClasses: Array<ExampleClass> = [];

// don't visit declare or remove nodes?
// no-undef uri, isStatic
declare module 'image!story-background' {
  declare var uri: string;
  declare var isStatic: boolean;
}
```

hzoo:
Ok so I guess we want `type x = any;` to act like a variable declaration? Ok so found that escope just has a `scope.variables` and looked for `scope.variables.push()`, which lead me to `currentScope().__define` which creates a variable.

From https://github.com/estools/escope/blob/master/src/variable.js#L72-L79 I found different variable types so I'm just using `Variable.Variable` which is just `"Variable"`

And then found out you can just do `referencer.prototype.TypeAlias` after removing the old code to not visit it to create a new variable definition!

Still need to add all the tests as well as visit the `TypeAlias.right`...

It looks like there is use of `type x = mixed` - is that supposed to be a built in type like `MixedTypeAnnotation`? @sebmck

sebmck:
@hzoo Nope. The `mixed in `type x = mixed;` is just a normal `GenericTypeAnnotation`.

hzoo:
Ok I guess there's just a lot of types that aren't imported explicitly in each file itself like `ReactComponent, ReactElement, ReactPropsCheckType, ReactPropsChainableTypeChecker, SyntheticEvent` in the `react-native` examples/libraries so eslint will give an error of `no-undef` (I'm assuming flow itself has all of these globally or from a file outside).

hzoo:
I'm just going to add tests for everything next - do you think I need add/change anything at the moment @sebmck?

sebmck:
Doesn't look too bad. Only comment would be on the use of `.bind(this)` which is a bit inefficient.

hzoo:
Yeah ok, I should replace all those will just regular for loops. And I guess update the PR based on the API changes you just did in #113?

Maybe the code would be better by refactoring it to use the `visitor-keys` data: https://github.com/babel/babel/blob/master/src/babel/types/visitor-keys.json#L79-L111?

Could something like this work? Might have to do some edge cases but it this seems to do the same as what I wrote befoe.

```js
var visitorKeysMap = {
  UnionTypeAnnotation: ["types"],
  TupleTypeAnnotation: ["types"],
  IntersectionTypeAnnotation: ["types"],
  TypeParameterDeclaration: ["params"],
  TypeParameterInstantiation: ["params"],
  TypeofTypeAnnotation: ["argument"],
  QualifiedTypeIdentifier: ["qualification"]
  ...
};

// how should we visit (a loop, a single property, check truthy)
var propertyTypes = {
  types: "loop",
  params: "loop",
  argument: "single", // or something
  qualification: "single",
  ...
};

function visitTypeAnnotation(node) {
  // get property to check (params, id, etc...)
  var visitorValues = visitorKeysMap[node.type];

  // can have multiple properties
  for (var i = 0; i < visitorValues.length; i++) {
    var visitorValue = visitorValues[i];
    var props = propertyTypes[visitorValue];
    if (props === "loop") {
      for (var j = 0; j < node[visitorValue]; j++) {
        checkIdentifierOrVisit.call(this, node[visitorValue][i]);
      }
    } else if (props === "single") {
      checkIdentifierOrVisit.call(this, node[visitorValue]);
    } else if (...) {
      ...
    }
  }
}
```

hzoo:
Ok refactored the `visitTypeAnnotation` method in https://gist.github.com/hzoo/15675611b9c7a65d3bb0. Ended up still being kind of long so don't know if that's a better idea or not since it's not as straightforward as before?

sebmck:
@hzoo Looks nicer to me, easier to add additional types later too.

hzoo:
Ok I'l go ahead and use that then. Is there a way to use `t.VISITOR_KEYS` but only for the flow types and ones with typeAnnotation? Or is it better to just make a copy like I already did.

sebmck:
I guess you could loop over and filter them. `require("babel-core").types.FLIPPED_ALIAS_KEYS.Flow` contains all the flow nodes.

hzoo:
Yeah don't know if it's that great

```js
  var visitorKeysMap = pick(t.VISITOR_KEYS, function(k) {
    return t.FLIPPED_ALIAS_KEYS.Flow.concat([
      "ArrayPattern",
      "ClassDeclaration",
      "ClassExpression",
      "FunctionDeclaration",
      "FunctionExpression",
      "Identifier",
      "ObjectPattern",
      "RestElement"
    ]).indexOf(k) === -1;
  });

  // don't check "id" to prevent "no-undef" for 'Component' with: 'let x: React.Component'
  visitorKeysMap.QualifiedTypeIdentifier = ["qualification"];
```

hzoo:
I added `Type Annotations`, `Array Types`, `Tuples` from the acorn flow tests https://gist.github.com/hzoo/c0ec600494025de9c192 and found/fixed a few issues because of them. Don't know if we need to do all of them but I can.  @sebmck 

For `react-native`, some variables aren't undefined so I added some to the `.eslintrc` globals:

```js
    "ReactClass": false,
    "ReactElement": false
```
Not sure about `x: mixed;`, `easing: ($Enum<typeof AnimationUtils.Defaults> | EasingFunction);`, or

```js
function renderApplication<D, P, S>(
  RootComponent: ReactClass<D, P, S>,
 ...
```
The only error in `Examples/` in `react-native` is `Example is defined but never used` https://github.com/facebook/react-native/blob/c700f71cb52362a80e7e740e85e09900058fbe27/Examples/UIExplorer/UIExplorerList.js#L33 since it doesn't seem to be used.

sebmck:
Unsure if they all needed to be added, just do whatever you feel comfortable with. Also, it might be best to reuse as much as you can from `babel.types.VISITOR_KEYS` so if extensions are done to the node types in the future then they'll automatically work in `babel-eslint`.

hzoo:
Ok I can add back something like the previous comment https://github.com/babel/babel-eslint/pull/109#issuecomment-105263162?

sebmck:
Sure, something like that would work.

hzoo:
Ok added `babel.types.VISITOR_KEYS`, removed more `.bind(this)`, and also trying to visit type casts
by removing change in https://github.com/babel/babel-eslint/issues/102 which fixes all the apparent issues I'm finding in `react-native`.

I think I'm ready to merge other than my confusion of if we need to do anything about `var x: mixed` since it will say it's not defined? @sebmck @wincent

sebmck:
> I think I'm ready to merge other than my confusion of if we need to do anything about `var x: mixed` since it will say it's not defined?

Seems fine to me. If this is actually an issue then it should be fixed by parsing it as a `MixedTypeAnnotation` node so there's nothing that needs to be handled on the `babel-eslint` side.

hzoo:
Ok good to merge then? And would this still be a patch version?

sebmck:
Looks GTM to me. Doesn't really matter what version it's released as since it's not breaking or "adding features" really.

tsing:
Hi, I tried the 3.1.10, it seems that it will got no-undef on polymorphic functions.

```
/* @flow */

export function foo<X>(x: X): X {
  return x;
}
``

the code above will got three no-undef errors:
```
  3:20  error  "X" is not defined  no-undef
  3:26  error  "X" is not defined  no-undef
  3:30  error  "X" is not defined  no-undef
```

hzoo:
I didn't figure out how to get polymorphic functions to work yet (stuff like `<T>`) - so, sorry I knew about those but still need to work on it.

tsing:
Could bable-eslint remove this `<T>` stuff such as it did in the previous versions until it can properly get the polymorphic functions  to work ?

hzoo:
The previous versions removed all flow types entirely but then you would get `(no-unused-vars)`.

Since they aren't going to refer to any types (`Foo`, `React`), we might as well remove them. Not sure how we can do this other than checking if the `Identifier` is a single capital letter (which really sounds like a hack) and deleting it? @sebmck

hzoo:
Ok @tsing see if 3.1.11 fixes things for you.

tsing:
@hzoo Thanks, it works well!

PR #114, opened 2015-05-25
support comprehensions (no-undef) - fixes #9

### LINKED ISSUE #9: Variable of array comprehension not defined
```js
let a = [], b = [for (e of a) String(e)];
```
results in:
```
  1:12  error  b is defined but never used  no-unused-vars
  1:22  error  'e' is not defined           no-undef
  1:37  error  'e' is not defined           no-undef
```

hzoo:
Fixes #9 
Only checking for `Identifier`: `(e of a)` and `ArrayPattern`: `for ([,x] of array)`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -276,6 +276,9 @@ function monkeypatch() {
         }
       }
     }
+    if (node.right) {
+      this.visit(node.right);
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


hzoo:
So this PR also triggers the `block-scoped-var` issue. (It's supposed to fix the `no-undef` issue). https://github.com/babel/babel-eslint/issues/9#issuecomment-105540883

```js
export default [for (i of global) String(i)];
```

I'm probably doing this all wrong haha. @michaelficarra What do you think we can do here?

Issue #62, opened 2015-04-06
ESLint failing tests


Cellule:
Hi, recently I have encountered quite a few problems with ESLint because I used babel-eslint. I would really like to keep using it for its better support for es6 and jsx.
I wanted to have an overview of the problems with babel-eslint so I executed ESLint test suite using babel-eslint and I wanted to share my results here.

The test suite found 115 failing test cases affecting 26 rules.
the rules are
- block-scoped-var
- brace-style
- comma-spacing
- complexity
- consistent-return
- default-case
- indent
- key-spacing
- no-catch-shadow
- no-div-regex
- no-empty-class
- no-ex-assign
- no-extra-parens
- no-fallthrough
- no-inline-comments
- no-multi-spaces
- no-regex-spaces
- no-shadow-restricted-names
- no-undefined
- padded-blocks
- space-before-blocks
- space-infix-ops
- spaced-line-comment
- valid-jsdoc
- vars-on-top
- wrap-regex

and you can see the detailed errors in the gist below (I enhanced a bit eslint-tester to have more information about failing tests)
https://gist.github.com/Cellule/37e1231437b1a28cdf46

I hope this can help solve most discrepancies between espree and babel-eslint.
If you want I can share how I ran the test suite since there is no official support at the moment.

sebmck:
> If you want I can share how I ran the test suite since there is no official support at the moment.

Sure, this would be great. Bit hard to ensure that these issues actually get fixed unless I can repeatedly run the test suite.

Cellule:
Right now I am working with eslint-tester to provide a way to change the config for all test cases https://github.com/eslint/eslint-tester/issues/19

However, in order to run the tests, I simply added a field on the ESlintTester prototype that would be used to set the parser for the tests.
Then I installed `babel-eslint` in the ESlint repo and I set the new ESlintTester field in `tests\lib\rules.js` to `babel-eslint` like
```js
var ESLintTester = require("eslint-tester");
ESLintTester.defaultParser = "babel-eslint";
```
Then every test ran after that point would use `babel-eslint` instead of `espree`

gabrielecirulli:
Please don't quote me on this, but might it be a change in the eslint api? I noticed this started to happen when eslint went from 0.16 to 0.17. We just locked the version to 0.16 and everything was fine, but then babel-eslint got update and the error got triggered again. It also happens on eslint 0.19.

gabrielecirulli:
[...]

PR #116, opened 2015-05-30
Add eslint tests - Ref #62

hzoo:
~~Not sure if this is the right way to do this (I think I did the submodule wrong)?~~
Also not sure about using `;` and `cd` in npm scripts and all that.
I did `git submodule add https://github.com/eslint/eslint eslint`

`npm run prepare-eslint-tests && npm run eslint-tests`

@sebmck, @Cellule

Getting 88 errors at the moment.


sebmck:
`cd` should be fine. `;` on Windows wont be though. Use `command && command` instead.

Cellule:
you could put `npm run prepare-eslint-tests && npm run eslint-tests` in an npm script or directly in script `test` like `mocha && npm run prepare-eslint-tests && npm run eslint-tests` I don't know if we want to run eslint tests everytime.

Also, I am not particularly familiar with git submodules, so I won't comment on it, but it's seems fine. I am only concern about publishing. Will the folder `eslint` be publish along with everything else ?

hzoo:
Ok fixed to use `&&` instead of `;`. Originally I put it in mocha but I figured it would be better to run it separately until we can get the tests to pass?
Also I don't think submodules are included in the publish? Looks like I need to add `git submodule update --init` like in babel which has traceur as a git submodule.

hzoo:
@sebmck So the eslint submodule won't be automatically be there when doing `npm install` since you have to run `git submodule update` right? So this should be good to run. And later we could probably remove the eslint devDependency and just use the submodule?

sebmck:
@hzoo You'll haev to add it to `.npmignore`.

hzoo:
Ah ok added @sebmck

hzoo:
Yeah a lot of errors are due to the tests using `with`.

```text
AssertionError: Should have no errors but had 1: [ { fatal: true,
    severity: 2,
    message: '\'with\' in strict mode',
```4

Also issue with tests involving jsx

```text
 6) comma-spacing <a>,</a>:
     Maximum call stack size exceeded
  ```

I guess we would need to change the parsing options for these tests only (loose mode)?

sebmck:
The `with` ones are alright since Babel is strict mode by default so those are correct. Second one is... weird.

hzoo:
Yeah really weird - seems to be failing at `escope.analyze` https://github.com/eslint/eslint/blob/master/lib/eslint.js#L666-L671

PR #117, opened 2015-06-01
update to eslint@0.22.1

hzoo:
Had to change the tests to fix the `no-unused-vars` error for classes since it looks like escope updated.


hzoo:
And if we go with #116, then we should be able to remove eslint from devDependencies and just use the git submodule?

PR #118, opened 2015-06-02
add flow exceptions for polymorphic types - Ref #109

hzoo:
For 

https://github.com/babel/babel-eslint/pull/109#issuecomment-107781203

```js
export function foo<X>(x: X): X {
  return x;
}
```
`  3:20  error  "X" is not defined  no-undef`

~~Also for `mixed` by not visiting those 2 cases.~~ Fixed in babel/babel#1669

### DIFF
--- a/index.js
+++ b/index.js
@@ -163,6 +163,10 @@ function monkeypatch() {
     if (node.typeAnnotation) {
       visitTypeAnnotation.call(this, node.typeAnnotation);
     } else if (node.type === "Identifier") {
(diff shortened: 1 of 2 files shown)


hzoo:
And since there are a lot of globals (`Array`), etc maybe we can import the definitions in https://github.com/facebook/flow/tree/master/lib core.js?

hzoo:
@sebmck this look good? There probably doesn't need to be a `PolymorphicTypeAnnotation` right?

hzoo:
good to merge?

PR #122, opened 2015-06-06
visit properties in ObjectPattern - fixes #95

### LINKED ISSUE #95: Properties excluded through spread + destructuring assignment should be allowed to be unused
Sample code is the easiest way for me to explain this. Say we have an object with a bunch of fields and you want to clone the object but exclude a couple of properties. We'd traditionally write:

``` js
var clone = Object.assign({}, originalObject);
delete clone.field1;
delete clone.field2;
```

With ES7 we can write:

``` js
var {field1, field2, ...clone} = originalObject;
```

Babel itself handles this correctly, but the linter complains if `field1` and `field2` are unused. Since the variable declarations have a meaningful side effect, I believe some code in the pipeline needs to tell eslint to ignore the unused variable declaration with `/*eslint-disable no-unused-vars */`, or perhaps Babel should omit the variable declaration if it lexically detects that the variable is used (although this may be undesirable if it breaks eval).

hzoo:


hzoo:
This should fix it @ide

AlexKVal:
I've checked it with that issue in `react-bootstrap`.
It works! :rainbow: :sparkle:

ide:
Thanks @hzoo!

hzoo:
Ok pushed 3.1.14!

AlexKVal:
Thanks @hzoo! :cherries:

PR #125, opened 2015-06-09
flow: generic/polymorphic type no-undef issue - fixes #123

### LINKED ISSUE #123: Flow generic class members
babel-eslint currently produces "unused" errors for valid flow-typed javascript when using generics with class members:

```js
function identity<T>(value: T): T {
  return value;
}
console.log(identity('foo'));

class Box<T> {
  value: T;
  constructor(value: T) {
    this.value = T;
  }
}
var box = new Box();
console.log(box.value);
```

Most uses of the generic seem to be fine, but typing the class members fails lint.

```
   7:9  error  "T" is not defined             no-undef
```

hzoo:
#123 
Hopefully that's all we need to do.. @leebyron

### DIFF
--- a/index.js
+++ b/index.js
@@ -158,15 +158,16 @@ function monkeypatch() {
     }
   }
 
-  function checkIdentifierOrVisit(node) {
+  function checkIdentifierOrVisit(node, typeParameters) {
     if (node.typeAnnotation) {
       visitTypeAnnotation.call(this, node.typeAnnotation);
     } else if (node.type === "Identifier") {
       // exception for polymorphic types: <T>, <A>, etc
-      if (node.name.length === 1 && node.name === node.name.toUpperCase()) {
+      if (node.name.length === 1 && node.name === node.name.toUpperCase() && typeParameters) {
         createScopeVariable.call(this, null, node);
+      } else {
+        this.visit(node);
       }
-      this.visit(node);
     } else {
       visitTypeAnnotation.call(this, node);
     }
@@ -184,12 +185,12 @@ function monkeypatch() {
     }
     if (node.typeParameters) {
       for (var j = 0; j < node.typeParameters.params.length; j++) {
-        checkIdentifierOrVisit.call(this, node.typeParameters.params[j]);
+        checkIdentifierOrVisit.call(this, node.typeParameters.params[j], true);
       }
     }
     if (node.superTypeParameters) {
       for (var k = 0; k < node.superTypeParameters.params.length; k++) {
-        checkIdentifierOrVisit.call(this, node.superTypeParameters.params[k]);
+        checkIdentifierOrVisit.call(this, node.superTypeParameters.params[k], true);
       }
     }
     visitClass.call(this, node);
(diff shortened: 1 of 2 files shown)


hzoo:
Ah currently it will also make a variable in the scope (`A`) when doing `var b: A = 1;` (probably isn't a big deal but might as error fix it to error correctly). That single line change is nice though :smile:

index.js
```diff
@@ -158,15 +158,16 @@ function monkeypatch() {
...
 
-  function checkIdentifierOrVisit(node) {
+  function checkIdentifierOrVisit(node, typeParameters) {
     if (node.typeAnnotation) {
       visitTypeAnnotation.call(this, node.typeAnnotation);
     } else if (node.type === "Identifier") {
       // exception for polymorphic types: <T>, <A>, etc
-      if (node.name.length === 1 && node.name === node.name.toUpperCase()) {
-        return;
+      if (node.name.length === 1 && node.name === node.name.toUpperCase() && typeParameters) {
```
leebyron: Do you still need to check if it's a one letter capital after this change? Not all type parameters are one letter capitals, so it would be nice if this new `typeParameters` bool sufficed :)

index.js
```diff
@@ -184,12 +185,12 @@ function monkeypatch() {
...
     if (node.typeParameters) {
       for (var j = 0; j < node.typeParameters.params.length; j++) {
-        checkIdentifierOrVisit.call(this, node.typeParameters.params[j]);
+        checkIdentifierOrVisit.call(this, node.typeParameters.params[j], true);
       }
     }
     if (node.superTypeParameters) {
       for (var k = 0; k < node.superTypeParameters.params.length; k++) {
-        checkIdentifierOrVisit.call(this, node.superTypeParameters.params[k]);
+        checkIdentifierOrVisit.call(this, node.superTypeParameters.params[k], true);
```
leebyron: superTypeParams should pass `, false` here, right? They're not defining a type param, but using one (and might actually use something else:

```
class A<T> extends B<T, number> { ... }
```

Here the `node.typeParameters` would be `T` which we should define in this new scope, and `node.superTypeParameters` would be `T, number` which we should check to see that they are defined in the scope.

test/non-regression.js
```diff
@@ -316,6 +316,53 @@ describe("verify", function () {
...
+            "var a: T = value; a;",
+          "};",
+          "var b: A = 1; b;",
+          "var c: T = 1; c;"
+        ].join("\n"),
+        { "no-unused-vars": 1, "no-undef": 1 },
+        [ '4:7 "A" is not defined. no-undef' ]
+      );
+    });
+
```
leebyron: should add a test like:

```
class A extends B<T> {}
```

which you would expect to complain that `T` is undefined

leebyron:
I took a stab at this in #126

hzoo:
K will close this in favor of #126.

index.js
hzoo: Not sure why but I was assuming that they were one capital letter given the examples I saw haha.

PR #126, opened 2015-06-09
Fixes flow type scoping issues

### LINKED ISSUE #123: Flow generic class members

leebyron:
This adds a special escope Scope when function, class, and typealias definitions include typeParameters which includes those parameters for reference, but is write-through to it's parent when encountering new variables. This is a little weird, but seems as good as we can while monkey-patching escope.

This also fixes some tests which had invalid flow but were written to expect no errors.

Fixes #123

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,29 +162,38 @@ function monkeypatch() {
     if (node.typeAnnotation) {
       visitTypeAnnotation.call(this, node.typeAnnotation);
     } else if (node.type === "Identifier") {
-      // exception for polymorphic types: <T>, <A>, etc
-      if (node.name.length === 1 && node.name === node.name.toUpperCase()) {
-        return;
-      }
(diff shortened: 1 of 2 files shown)


hzoo:
Yeah I think I definitely misunderstood how the generic types work but it makes sense now.

hzoo:
Ok will merge and release a new version @leebyron.

Issue #124, opened 2015-06-08
`space-in-parens` error for code in a comment block


necolas:
This looks like a bug because it's reporting an error on a commented-out line.

Error:

```
  4:6  error  There should be no spaces inside this paren  space-in-parens
```

Source:

```js
 render() {
    // return (
    //   <div />
    // ); // <-- this is the line that is reported
  }
```

hzoo:
@necolas Looks like the acorn/espree ast is different regarding comments, although not sure why that affects that rule.. - will investigate.

necolas:
Thanks. And FYI, switching to a wrapping `/* */` comment avoids the false positive.

sebmck:
@hzoo Try filtering out `CommentBlock` and `CommentLine` tokens.

hzoo:
Kind of confused - and it looks like if it's not empty it also works?
```js
 render() {
    1;
    // return (
    //   <div />
    // ); // <-- this is the line that is reported
  }
```
@sebmck Sorry how would I do that? It looks like comments and tokens are in separate arrays? Sounds like a similar issue to https://github.com/babel/babel/issues/672 of using estraverse to attach comments to empty block `estraverse.attachComments(ast, comments, tokens);`? (Although this works with regular eslint so maybe not)?

hzoo:
Ok so this is related to https://github.com/babel/babel-eslint/pull/76#issuecomment-95360914? I guess comments aren't attached in an empty block, which could be solved but not using estraverse like the `Attach comments in parser` task in https://github.com/babel/babel/issues/1486

hzoo:
I guess this can be closed as a upstream babel issue actually

hzoo:
Ok looks like this was also fixed in the other comment changes https://github.com/babel/acorn-to-esprima/commit/5da46d46e842897f5e3b261303994273e0933064

hzoo:
@necolas let me know if it's still an issue

PR #128, opened 2015-06-10
add tests - for #124

hzoo:
#124

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -1238,4 +1238,38 @@ describe("verify", function () {
       []
     );
   });
+
+  // it("line comment space-in-parens #124", function () {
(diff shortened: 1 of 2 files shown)


test/non-regression.js
```diff
@@ -1135,4 +1135,38 @@ describe("verify", function () {
...
   });
+
+  it("line comment space-in-parens #124", function () {
+    verifyAndAssertMessages(
+      [
+        "React.createClass({",
+          "render() {",
+             "// return (",
+             "//   <div />",
+             "// ); // <-- this is the line that is reported",
```
hzoo: Looks like this particular test still fails after the comment changes in babel because of `//_` where _ is the space. Not sure why yet - tokens and comments look fine.

PR #135, opened 2015-06-23
Fix eslint module path resolution in jest

PiPeep:
Jest doesn't implement module.parent (as per facebook/jest#411), so this falls back to `require.resolve` if the `Module._resolveFilename` call fails.

See https://github.com/pipeep/babel-eslint-jest-testcase for an example testcase.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,9 +22,15 @@ function monkeypatch() {
 
   var eslintLoc;
   try {
+    // avoid importing a local copy of eslint, try to find a peer dependency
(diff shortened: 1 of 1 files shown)


hzoo:
looks good to me - @sebmck?

sebmck:
:+1:

hzoo:
Thanks @PiPeep for PR and explanation!

PiPeep:
Thanks! This should save me from ripping the rest of my hair out.

PR #143, opened 2015-07-05
only visit properties in object destructuring when there is a SpreadP…

### LINKED ISSUE #142: no-unused-vars is not reporting for object destructuring
No error:
```js
const {connect} = require('redux/react')
```

Error
`error  no-unused-vars  connect is defined but never used`
```
const connect = require('redux/react').connect
```

hzoo:
…roperty - fixes #142 

WIP

@sebmck How can I check if one of the properties is a spread property (since we changed the `node.type`). I just changed it to SpreadProperty but I'm assuming that may break something?

Since we want to only visit when a SpreadProperty is used (and I guess more specifically visit only properties left of it)

Ref #95 
@cassus tell me if this fixes your issue

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -172,7 +172,7 @@ var astTransformVisitor = {
   noScope: true,
   exit: function (node) { /* parent */
     if (this.isSpreadProperty()) {
-      node.type = "Property";
+      node.type = "SpreadProperty";
       node.kind = "init";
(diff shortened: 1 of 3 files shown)


sebmck:
@hzoo IIRC, some ESLint rules would complain but I think that was just about the properties and not the actual node type so if this passes some/all the ESLint tests then it should be right.

PR #146, opened 2015-07-06
use node.babelType for SpreadProperty

hzoo:
Is this good?

@sebmck @jquense


sebmck:
I meant more putting `node._babelType = node.type;` at the top of the `enter` visitor so all nodes have their original pure type.

hzoo:
Ah ok that makes more sense

hzoo:
this what you meant?

sebmck:
:+1:

PR #152, opened 2015-07-21
Replace 'Line X' with actual line number

bgw:
This makes the output closer to espree and less confusing.

@sebmck, you introduced the code in fc8dfe36e7be7fa508f480dbd76822f6abc30552, which makes it look like this was intentional. Am I missing something?

### DIFF
--- a/index.js
+++ b/index.js
@@ -432,7 +432,7 @@ exports.parse = function (code) {
       err.column = err.loc.column;
 
       // remove trailing "(LINE:COLUMN)" acorn message and add in esprima syntax error message start
-      err.message = "Line X: " + err.message.replace(/ \((\d+):(\d+)\)$/, "");
+      err.message = "Line " + err.lineNumber + ": " + err.message.replace(/ \((\d+):(\d+)\)$/, "");
     }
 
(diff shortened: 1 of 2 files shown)


sebmck:
@bgw ESLint would require the `message` to follow this format or the regex match it used would fail, this may have changed in later versions.

bgw:
@sebmck Thanks for the quick reply. It seems it doesn't break anything. I applied this patch in react, messed up a random file, and checked the output. The top is with my patch. The bottom is without:

screen shot 2015-07-20 at 6 06 42 pm

sebmck:
I think this was changed via eslint/eslint@7a3c545272821a2aa47bfd87b0c01ba48812616f. Will likely require a major version bump of `babel-eslint` to prevent breaking people using older versions of ESLint.

hzoo:
Looks good? Should we ahead and merge and do 4.0.0?

sebmck:
Yeah, if the tests pass then :+1:

hzoo:
Ok thanks @bgw publishing 4.0.0 now

Issue #11, opened 2015-02-28
rest parameter for destructuring assignments results as undefined variable


mistadikay:
For code like this:
```js
const [a, ...rest] = ['1', '2', '3'];
console.log(a);
console.log(rest);
```

I'm having an error:
```bash
13:12  error    'rest' is not defined                   no-undef
```

sebmck:
Fixed as of 1.0.10, thanks!

PR #164, opened 2015-08-15
Remove rest transform

hzoo:
Ref #11 

escope issue fixed in 3.0.0 https://github.com/estools/escope/issues/47


hzoo:
@sebmck look good?

hzoo:
Ok merging this first since should be least likely to have issues.

Issue #7, opened 2015-02-27
"is not defined" for ...rest arguments


gaearon:
```js
function createStoreMixin(...stores) {
  var StoreMixin = {
    componentDidMount() {
      this.stores = stores;
```

will say `stores is not defined`. This happens in any method with `...rest` arguments.

sebmck:
Fixed as of 1.0.8, thanks!

gaearon:
Thank you!

kushalmahajan:
I'm still getting the error

```
export const rootEpic = (action$, state$) => epic$.pipe(
    mergeMap(epic => 
        epic(action$, ...rest).pipe(
            takeUntil(action$.pipe(
            ofType('EPIC_END')
            ))
        )
        // epic(action$, state$)
    )
);
```

nicolo-ribaudo:
Where are you defining `rest`?

nicolo-ribaudo:
There `rest` is defined as a parameter of the outer arrow function; in your example it isn't.

PR #165, opened 2015-08-15
use SpreadProperty type, revert other transforms

hzoo:
Ref original issue https://github.com/babel/babel-eslint/issues/7

`node.key = node.value = node.argument;` is needed (failed test)

Fixes xjamundx/eslint-plugin-standard#3

Looks like no issue with babel-eslint, react, react-bootstrap, express-graphql, or graphql-js tests.

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -183,11 +183,8 @@ var astTransformVisitor = {
   },
   exit: function (node) { /* parent */
     if (this.isSpreadProperty()) {
-      node.type = "Property";
-      node.kind = "init";
-      node.computed = true;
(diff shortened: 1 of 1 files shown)


hzoo:
This would mean you wouldn't have to use `node._babelType` for the SpreadProperty @jquense (assuming it didn't break again)

jquense:
is this possible because of spread support in eslint 1.0? I was under the impression that they were calling that node type "ExperimentalSpreadProperty" or is this a different thing completely?

hzoo:
No, I choose to ignore that - I just removed what was unnecessary as long as the tests still passed and then ran it on a few repos (listed above).

Issue #168, opened 2015-08-16
space-before-function-paren and async function


jupl:
The following line:

```js
it('handles updates', async function() {
                           ^
```

Is throwing up a flag under the caret with the violation of `space-before-function-paren`. This occurs in v4.0.7 and has not thrown an error in previous versions.

hzoo:
Thanks for the report - I'l try to fix this soon. It's definitely because of https://github.com/babel/babel-eslint/pull/166. Need to fix this and the other issue somehow - but this is a good test to add in.

hzoo:
Ok should be fixed in `4.0.8`

taion:
I'm still seeing issues here against `v4.0.10`. For:

```js
async function foo() {
  return 5;
}

export default foo;
```

I get

```
test.js
  1:15  error  Unexpected space before function parentheses  space-before-function-paren
```

hzoo:
Dang - but hmm cannot reproduce? on eslint@1.1.0 as well.

```
//.eslintrc
parser: "babel-eslint"
rules:
    space-before-function-paren: [2, "never"]
```

taion:
Sorry - something terribly messed up in my environment. It went away.

PR #169, opened 2015-08-16
async func and space-before-function-paren

hzoo:
For #168.

WIP: Probably not the fix we want but added the test.

Actually based on https://github.com/eslint/eslint/blob/master/lib/rules/space-before-function-paren.js#L71-L73 maybe it is. Other solution would be to remove the transform and that the rule to eslint-plugin-babel with a check for `node.async`. Or we do the private properties thing again like `._babelType`...

### DIFF
--- a/acorn-to-esprima.js
+++ b/acorn-to-esprima.js
@@ -235,6 +235,10 @@ var astTransformVisitor = {
 
     // functions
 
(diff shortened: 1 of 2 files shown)


hzoo:
If still there's issues after this then the rules should probably be rewritten in `eslint-plugin-babel`

PR #202, opened 2015-11-03
update to babel6

hzoo:
related PR https://github.com/babel/acorn-to-esprima/pull/3

@sebmck 

- specific babel packages
- remove references to comprehensions
- update acorn-to-esprima
- update babylon options

Need to add more test cases/repos to test


lyonlai:
any update on this?

olalonde:
Bump... had to disable eslint after switching to babel 6.

reggi:
What's the word on this?

evocateur:
there are still a few issues I've discovered, deep in the nasty bowels of babylon as far as I can tell. will report back after my spelunking is finished.

hzoo:
There is definetely an issue with the loc/range for objectproperty/objectmethod changes

evocateur:
I've been hitting issues with `StringLiteral` nodes losing their `extra` property (and thus their derived `raw` property, which eslint uses in various checks, like `jsx-quotes`). A `StringLiteral` node's `extra` property is being [set to `null` in babylon](https://github.com/babel/babel/blob/master/packages/babylon/src/plugins/jsx/index.js#L249), and I don't understand why the original metadata isn't allowed through.

hzoo:
- [ ] ClassMethod fails
```js
class A {
  constructor(a) {

  }
}
```

hzoo:
Going to do a prerelease

Issue #204, opened 2015-11-06
Getting 'no-undef' warnings when using decorators.


martndemus:
When using eslint-babel, I get `no-undef` warnings when declaring decorators on object properties.

```javascript 
 var Model = DS.Model.extend({
    @attr firstName,
    @attr({ defaultTo: 'blue' }) lastName,
    @hasMany user,
    @belongsTo car
  });
```

```
  14:11  error  "firstName" is not defined  no-undef
  15:34  error  "lastName" is not defined   no-undef
  16:14  error  "user" is not defined       no-undef
  17:16  error  "car" is not defined        no-undef
```

Setup: 
```
$ npm ls | grep eslint
├─┬ ember-cli-qunit-eslint@1.0.0
│ ├─┬ babel-eslint@4.1.4
│ ├─┬ broccoli-lint-eslint@1.1.1
│ │ └─┬ eslint@1.8.0
```

```
Some of the output has been trimmed here.
$ npm ls | grep babel
├─┬ ember-cli-babel@5.1.5
│ ├─┬ broccoli-babel-transpiler@5.4.5
│ │ ├─┬ babel-core@5.8.33
│ ├─┬ babel-eslint@4.1.4
│ │ ├─┬ babel-core@5.8.33
```

martndemus:
Hmm, I'm doubting this is a real bug. It might actually be a valid warning.

Example: 
```javascript
{
  @readOnly foo
}
```

is transpiled by babel to

```javascript
{
  key: 'foo',
  decorators: [readOnly],
  initializer: function initializer() {
    return foo;
  }
}
```

Which looks like it's just like a shorthand property.

```javascript
var a = 1;
var b = 2;

var c = { a, b }

c.a // => 1
c.b // => 2
```
So this would cause a nice subtle bug if we would disable `no-undef` warnings on this.
```javascript
var foo = 'bar'

var object = {
  @readOnly foo
};

object.foo // => 'bar'
```

What do you think? @hzoo

hzoo:
Right those are shorthand props - I guess I assumed those were defined somewhere else or were parameters. If the code snippet is just

```
var Model = DS.Model.extend({
    @attr firstName,
    @attr({ defaultTo: 'blue' }) lastName,
    @hasMany user,
    @belongsTo car
  });
```
Then yeah I guess it should error since it would error even without the decorators

devinus:
[...]

PR #205, opened 2015-11-06
Failing test for issue #204

martndemus:
I'll take a shot at fixing the issue. Help very wanted.

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -1219,6 +1219,24 @@ describe("verify", function () {
     });
   });
 
+  it("detects no no-undef on decorated property #204", function () {
+    verifyAndAssertMessages(
+      [
+        "import classMethodDeclarationA from 'decorator';",
+        "import decoratorParameter from 'decorator';",
+        "var obj = {",
+          "@classMethodDeclarationA(decoratorParameter)",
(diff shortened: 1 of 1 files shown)


hzoo:
`propertyDeclaration` is not undefined. What we're doing is pretty insane but yeah. Not sure

You could heck the method `visitDecorators` https://github.com/babel/babel-eslint/blob/master/index.js#L88-L97 and the call sites - although that is used to fix`no-unused-vars` (the corresponding rule)

martndemus:
Hmm, I'm doubting this is a real bug. It might actually be a valid warning.

Example: 
```javascript
{
  @readOnly foo
}
```

is transpiled by babel to

```javascript
{
  key: 'foo',
  decorators: [readOnly],
  initializer: function initializer() {
    return foo;
  }
}
```

Which looks like it's just like a shorthand property.

```javascript
var a = 1;
var b = 2;

var c = { a, b }

c.a // => 1
c.b // => 2
```
So this would cause a nice subtle bug if we would disable `no-undef` warnings on this.
```javascript
var foo = 'bar'

var object = {
  @readOnly foo
};

object.foo // => 'bar'
```

PR #209, opened 2015-11-20
Patch SpreadProperty to escope's PatternVisitor

Constellation:
Once new escope is released, we can patch SpreadProperty to PatternVisitor.
This will solve estools/escope#88 and eslint/eslint#4481.

See escope side PR: https://github.com/estools/escope/pull/89

PR #212, opened 2015-11-23
Avoid errors with eslint 1.10.x when excluding destructured properties

evocateur:
Despite @Constellation's improvements from #209 with the pattern visitor, it doesn't actually iterate over the Property node siblings of the SpreadProperty. Thus, the "ignore properties left of spread" fix in #95 regressed.

Restoring the "old" path of looping over all properties of an ObjectPattern that contains a SpreadProperty avoids the regression. It seems inelegant, but I don't have enough insight into the referencer to propose an alternative.

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,18 +295,15 @@ function monkeypatch() {
         if (typeAnnotation) {
           checkIdentifierOrVisit.call(this, typeAnnotation);
         }
-        if (!patternVisitor) {
-          // Old method. Once escope in eslint is updated, this code is not necessary.
-          if (id.type === "ObjectPattern") {
-            // check if object destructuring has a spread
-            var hasSpread = id.properties.filter(function(p) {
-              return p._babelType === "SpreadProperty";
-            });
-            // visit properties if so
-            if (hasSpread.length > 0) {
-              for (var j = 0; j < id.properties.length; j++) {
-                this.visit(id.properties[j]);
-              }
+        if (id.type === "ObjectPattern") {
+          // check if object destructuring has a spread
+          var hasSpread = id.properties.filter(function(p) {
+            return p._babelType === "SpreadProperty";
+          });
+          // visit properties if so
+          if (hasSpread.length > 0) {
+            for (var j = 0; j < id.properties.length; j++) {
+              this.visit(id.properties[j]);
             }
           }
         }


hzoo:
Hmm so I merged https://github.com/babel/babel-eslint/pull/209 (and we see the tests fail) - but I didn't publish a new version yet. Maybe we should just revert the change then? And does that mean 4.1.5 is already working with 1.10.x since it doesn't seem like it?

evocateur:
@hzoo Nope, sadly, both #209 and this patch are necessary. With v4.1.5, the spread properties are still erroring with the `no-redeclare` error.

evocateur:
That is to say, with the latest eslint (1.10.x). If you're lucky enough to be shrinkwrapped to eslint 1.9.0 with escope 3.2.x, then you're unbroken.

hzoo:
Oh ok. I guess I'm confused as to why we need it since if the spreadproperty isn't visited from the previous patch and this one just uses the old code path, wouldn't this be the same as reverting?

evocateur:
Oh, no, the SpreadProperty visitor is still used. For reasons I lack the parser experience to grok, when you *only* visit the SpreadProperty, the leftover Property siblings cause trouble.

evocateur:
I think because without the SpreadProperty patched onto escope's referencer, it will use the Property visitor, which is not sufficient for the purposes of babel-eslint.

evocateur:
And really, #95 is such a weird change, to my naive mind, looks "wrong". If I destructured those identifiers and then didn't use them, I would *expect* `no-unused` to raise an error. (I read all the comments three times, and I guess I disagree with the spec; if you don't want to use properties of the copied object, why not just `delete` them?)

hzoo:
Ok cool. I'l prob just merge this and look into it later to see if we have a better solution. Well one issue is in order to use `var { a, ...rest} = obj` you need to specify which properties you don't want from `rest`. In this sense they are used to define that will be in `rest` and thus used?

index.js
```diff
@@ -295,18 +295,15 @@ function monkeypatch() {
...
-              }
+        if (id.type === "ObjectPattern") {
+          // check if object destructuring has a spread
+          var hasSpread = id.properties.filter(function(p) {
+            return p._babelType === "SpreadProperty";
+          });
+          // visit properties if so
+          if (hasSpread.length > 0) {
+            for (var j = 0; j < id.properties.length; j++) {
+              this.visit(id.properties[j]);
```
hzoo: Oh yeah, I see I made it so we visit all properties (both Property and SpreadProperty)

evocateur:
I'm new enough to ES6 syntax, generally, that the spec-compliant behaviour in this case just feels wrong to me. Yes, it is correct that you need to specify what you don't want from an object before the spread; my own personal preference is to be explicit about the things I intend to use (specified identifiers) vs those that I don't (deleted properties).

Anyway, thanks for being so responsive today as I struggle to understand complex topics!

index.js
evocateur: I experimented with filtering only non-SpreadProperty types (`_babelType === "Property"`), but I felt that was more error-prone.

hzoo:
Haha no problem - I'm a noob at all this as well. Also to clarify - this isn't really spec compliant in the since that its for sure since object rest spread is in stage 2 as of right now (https://github.com/tc39/ecma262#current-proposals).

So it's not actually ES6 - there is SpreadElement and RestElement (used in params and arrays like `function a (...params) {}` and `var a = [...arr, ...arr2];` This is for objects/object destructuring https://github.com/sebmarkbage/ecmascript-rest-spread

evocateur:
Ah, good to know!

PR #264, opened 2016-02-28
ESLint 2

hzoo:
Sigh, still a lot of failing tests, I just updated the eslint/espree options to make 50 tests pass but there's still like 20 related to the acorn/espree 3.0 change. Not fun..

Will need to be fixed in the acorn-to-esprima package - although at this point that doesn't make sense at all :sweat_smile:

### DIFF
--- a/babylon-to-espree/attachComments.js
+++ b/babylon-to-espree/attachComments.js
@@ -1,0 +1,57 @@
+// comment fixes
+module.exports = function (ast, comments, tokens) {
+  if (comments.length) {
+    var firstComment = comments[0];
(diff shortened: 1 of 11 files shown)


danez:
Thanks for taking care of this.
This snippet is currently still not working with eslint2 and your branch:
```
const Paragraph = props => {
  const { children, ...rest } = props;

  return (
    <p {...rest}>{children}</p>
  );
};
```

`line 2: children was used before it was defined [no-use-before-define]`

hzoo:
Ok since tests pass we can release a beta version for this.

PR #269, opened 2016-03-07
try catch estraverse-fb since eslint removed it in 2.3.0, temp skip b…

hzoo:
…roken tests until we come up with a solution

Ref https://github.com/eslint/eslint/issues/5476

babel-eslint monkey patches the visitorKeys of eslint for things like flow, jsx, experimental syntax (async, decorators, etc) but in ESLint 2.3.0 certain things were changed/removed.

Mostly that estraverse-fb was removed, https://github.com/eslint/eslint/pull/5443/ so we can just check for that

https://github.com/eslint/eslint/pull/5478, https://github.com/eslint/eslint/issues/4640 should help

### DIFF
--- a/index.js
+++ b/index.js
@@ -55,10 +55,12 @@ function monkeypatch() {
   estraverses.push(estraverse);
   assign(estraverse.VisitorKeys, t.VISITOR_KEYS);
 
-  // monkeypatch estraverse-fb
-  var estraverseFb = eslintMod.require("estraverse-fb");
-  estraverses.push(estraverseFb);
-  assign(estraverseFb.VisitorKeys, t.VISITOR_KEYS);
+  // monkeypatch estraverse-fb (only for eslint < 2.3.0)
+  try {
+    var estraverseFb = eslintMod.require("estraverse-fb");
+    estraverses.push(estraverseFb);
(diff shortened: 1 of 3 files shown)


tupton:
Is there any info on when this will be released? Would it be in a 5.0.x release or would we have to update to a 6.0 beta to get this fix?

hzoo:
This actually doesn't fix the probelm since the visitorkeys themselves aren't overridable in ESLint 2.3.0.

I did release a 6.0.0-beta.2 with this fix but it's better to pin to ESLint 2.2.x since there are other issues

tupton:
Thanks for the update.

Unfortunately, I'm using gulp-eslint, which probably doesn't want to pin to 2.2.x just for babel-eslint. The other "workaround" involves installing `estraverse-fb` explicitly, which I don't think has any other side-effects and seems to be getting the job done for now.

I'll keep an eye on those other eslint issues you mentioned above, too. Thanks again for the great work.

hzoo:
If it's working for you then maybe I can do a 5.0.1 with this - but I saw https://github.com/eslint/eslint/issues/5494 and others (internal rules erroring as a result).

tupton:
Ah, good point; I didn't see those other issues. I'll see if there's a better way to pin to 2.2.x. Thanks!

PR #270, opened 2016-03-07
remove monkey-patching of patternVisitor as unused

danez:
SpreadProperty will never be visited as it is now called ExperimentalSpreadProperty
We also don't need to fix it to the correct node-type, as eslint itself takes care about ExperimentalSpreadProperty


hzoo:
Great catch thanks!

PR #271, opened 2016-03-07
Remove visiting of properties left of spread, fixes #249.

### LINKED ISSUE #249: no-used-before-define is erroneously throw when destructuring object with rest operator
I think this one is legit :)
### How to reproduce
Using `babel-eslint@5.0.0-beta10`
**.eslintrc**
```json
{
    "parser": "babel-eslint",
    "env": {
        "es6": true,
    },
    "parserOptions": {
        "ecmaFeatures": {
            "experimentalObjectRestSpread": true
        },
    },
    "rules": {
        "no-use-before-define": 2,
    }
}
```
**test.js**
```js
const { a, ...b } = { a: 1, b: 2, c:3 };
console.log(a, b);
```

When executing
```shell
$ eslint test.js
```
The linter raise the following error
```
test.js
   2:9   error  'a' was used before it was defined  no-use-before-define
   2:15  error  'b' was used before it was defined  no-use-before-define
```
(nothing appear when using the eslint parser)

danez:
I was trying hard to get both #95 and #249 working together, but I couldn't find a way that does not involve cheating by changing line/column numbers of nodes.

Maybe the best solution for now is to revert #95 and make sure that at least the core eslint stuff is working (and also eslint with espree is no doing the "visit properties left of spread"). Reverting #95 is what I did in this PR.

If someone finds a way of doing #95 without breaking #249 again, it can also be done later, but I think it is more important to make sure all rules work as expected.

### DIFF
--- a/index.js
+++ b/index.js
@@ -320,18 +320,6 @@ function monkeypatch() {
         if (typeAnnotation) {
           checkIdentifierOrVisit.call(this, typeAnnotation);
         }
-        if (id.type === "ObjectPattern") {
-          // check if object destructuring has a spread
-          var hasSpread = id.properties.filter(function(p) {
-            return p._babelType === "SpreadProperty" || p._babelType === "RestProperty";
-          });
-          // visit properties if so
-          if (hasSpread.length > 0) {
-            for (var j = 0; j < id.properties.length; j++) {
-              this.visit(id.properties[j]);
-            }
-          }
-        }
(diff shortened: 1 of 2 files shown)


hzoo:
@ide Thoughts?

ide:
Reverting looks like the lesser of two evils but we should re-open #95 then.

PR #273, opened 2016-03-10
Fix: Improve monkeypatching for ESLint v2.3.0

mysticatea:
Since v2.3.0, ESLint came to use `keys` and `childVisitorKeys` options instead of extending `estraverse. VisitorKeys`.
ESLint team has the plan to add the ability which extends VisitorKeys into parser plugins, but it may take a long time.

So this PR improves monkey patching for those options.

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,7 @@ function monkeypatch() {
     estraverses.push(estraverseFb);
     assign(estraverseFb.VisitorKeys, t.VISITOR_KEYS);
   } catch (err) {
-    throw new Error("babel-eslint isn't currently compatible with ESLint 2.3.x. The recommendation is to pin to ESLint 2.2.x right now.");
+      // Ignore: ESLint v2.3.0 does not have estraverse-fb
   }
 
   // ESLint v1.9.0 uses estraverse directly to work around https://github.com/npm/npm/issues/9663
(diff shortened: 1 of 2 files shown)


hzoo:
Thanks @mysticatea!

I added in https://github.com/babel/babel-eslint/blob/32848e57bd606076c4e0c2a7c618ff6a6ab639ff/index.js#L353 - is that not necessary anymore?

Also for this PR, I would bump https://github.com/babel/babel-eslint/blob/master/package.json#L36 to 2.3.0 so it will actually run on it

mysticatea:
Ouch, I updated https://github.com/babel/babel-eslint/blob/master/package.json#L36 to 2.3.0, then 1 test came to failing.
I'm investigating it.

hzoo:
Ah ok, it might not be due to your change specifically though

mysticatea:
Anyway, I share it.

mysticatea:
I'm sorry, I don't have enough time to investigate it today (JST)...

hzoo:
Thanks - maybe I can check it out tonight

hzoo:
Reran the tests-

mysticatea:
@nzakas is faster than me.

- That test failing is caused by [here](https://github.com/babel/babel-eslint/blob/master/index.js#L278-L287). This patching is changing traversing properties dynamically. But it cannot be changed dynamically in my way. 
- `babel-eslint` seems to work with `eslint@2.4.0`.

So I just remove throwing an error when `estraverse-fb` has not found.

PR #282, opened 2016-03-30
Fix processing sourceType: script

josh:
It seems like some recent version of babel-eslint broke the [no-implicit-globals](http://eslint.org/docs/rules/no-implicit-globals) rule.

I'm not sure what specific version, but I'll try to bisect to figure out when it happened.

That test case can be verified working when `parser: "babel-eslint"` is commented out and broken once enabled.

I've only dug in a little, but it seems like its an issue with the scope analysis. `context.getScope().variables` doesn't seem to be reporting the correct global variables in the script type case.

Thanks!

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,7 +76,6 @@ function monkeypatch() {
   var analyze = escope.analyze;
   escope.analyze = function (ast, opts) {
     opts.ecmaVersion = 6;
-    opts.sourceType = "module";
 
     var results = analyze.call(this, ast, opts);
     return results;
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -2,27 +2,32 @@
 "use strict";
 var eslint = require("eslint");
 
-function verifyAndAssertMessages(code, rules, expectedMessages, sourceType) {
-  var messages = eslint.linter.verify(
-    code,
-    {
-      parser: require.resolve(".."),
-      rules: rules,
-      env: {
-        node: true,
-        es6: true
+function verifyAndAssertMessages(code, rules, expectedMessages, sourceType, overrideConfig) {
+  var config = {
+    parser: require.resolve(".."),
+    rules: rules,
+    env: {
+      node: true,
+      es6: true
+    },
+    parserOptions: {
+      ecmaVersion: 6,
+      ecmaFeatures: {
+        jsx: true,
+        experimentalObjectRestSpread: true,
+        globalReturn: true
       },
-      parserOptions: {
-        ecmaVersion: 6,
-        ecmaFeatures: {
-          jsx: true,
-          experimentalObjectRestSpread: true,
-          globalReturn: true
-        },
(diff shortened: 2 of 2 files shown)


josh:
Okay, maybe I figured it out. Looks like `sourceType = "module"` was hard coded in the `escope.analyze` monkey patch.

hzoo:
Hmm yeah I think we assumed that since it's es2015 it would all be modules. I guess we acutally need to be able to get eslintrc options and pass them into here.

josh:
> Hmm yeah I think we assumed that since it's es2015 it would all be modules.

Ah, yeah, we're definitely using it on some *simple* scripts.

> I guess we acutally need to be able to get eslintrc options and pass them into here.

Looks like thats the case, `opts` is passed along from eslintrc options.

If you think this is going to break something else, are there any other cases I can add to the test suite here?

hzoo:
Oh it is being passed in now? I wasn't paying attention to when that would happen. We can just read from that in order to set stuff like that then.

josh:
A `console.log` dump within that call:

``` js
{ ignoreEval: true,
  nodejsScope: undefined,
  impliedStrict: undefined,
  ecmaVersion: 5,
  sourceType: 'script',
  fallback: [Function] }
```

``` js
{ ignoreEval: true,
  nodejsScope: false,
  impliedStrict: undefined,
  ecmaVersion: 6,
  sourceType: 'module',
  fallback: [Function] }
```

hzoo:
Or... I'm just confusing myself since we already are doing it lol.

josh:
I wonder if defaulting to `"module"` here is *okay* too

https://github.com/babel/babel-eslint/blob/636850bea94e72783c587fd018e724362493a1c3/index.js#L369

josh:
Anyway, if you think activating the parse should put you in module mode thats fine with me as along as explicitly setting a script type still works (which is whats busted right now).

hzoo:
I was thinking the options passed into parse from eslint might be useful

josh:
Great! Is there anything else you'd like me to change or does this look fine?

hzoo:
Hm I don't think so - if the sourcetype changes based on the options then it makes sense rather than the hardcoding.

hzoo:
Btw, what version are you using? There is a 5.x branch which is for eslint1 and a 6.x branch (master) which is for eslint 2. Might have to port the fix to 5.x

josh:
> if the sourcetype changes based on the options then it makes sense rather than the hardcoding.

Yep, thats correct.

> Btw, what version are you using?

I've been testing against eslint 2.x.

I'll see about opening another PR to backport the fix for eslint 1.x

Thanks!

josh:
Posted the backport in https://github.com/babel/babel-eslint/pull/283

hzoo:
Alright thanks for sticking through with this - published 6.0.1 and 5.0.3

PR #287, opened 2016-03-31
Restore "Fix processing sourceType: script"

josh:
NOT READY TO BE MERGE

-
CC: @hzoo
Reverts babel/babel-eslint#285

### DIFF
--- a/index.js
+++ b/index.js
@@ -76,7 +76,6 @@ function monkeypatch() {
   var analyze = escope.analyze;
   escope.analyze = function (ast, opts) {
     opts.ecmaVersion = 6;
-    opts.sourceType = "module";
 
     var results = analyze.call(this, ast, opts);
     return results;
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -2,27 +2,32 @@
 "use strict";
 var eslint = require("eslint");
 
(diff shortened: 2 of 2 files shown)


josh:
Need to figure out how to default to `scriptType: "module"` while allowing the option to be overriding by eslint configuration.

hzoo:
So I think we pass the options from parse into monkeypatch and then pass that into the escope part and do `options.sourceType || "module"`

josh:
> So I think we pass the options from parse into monkeypatch and then pass that into the escope part and do options.sourceType || "module"

I don't think that will work. `opts.sourceType` is always set. See it initialized here in eslint  https://github.com/eslint/eslint/blob/2bb8486380a74b076b0fb3244310d0a9edf12d4f/lib/eslint.js#L837

Maybe if we could somehow access `currentConfig.parserOptions.sourceType`?

danez:
I think it should work, as the options that the parser gets are the raw options from eslint (without defaults) + some stuff that eslint needs (loc, range, ...)
I created a PR that should work, can you guys have a look? #288

josh:
Closing this in favor of https://github.com/babel/babel-eslint/pull/288

Issue #217, opened 2015-11-30
[rule:no-redeclare] not working in babel-eslint


anshulguleria:
I am using `eslint v1.9.0`, `node v5.0.0` and `babel-eslint v4.1.5`.

I was trying to use `no-redeclare` rule in `eslintrc` file.
```json
{
    "parser": "babel-eslint",
    "rules": {
        "no-redeclare": [2, { "builtinGlobals": true }]
    }
}
```
And code as:
```js
let mine = 23;
let mine = 34;

console.log(mine);
```

Ideally this should give me `no-redeclare` error but it didn't.

When I removed the `parser` configuration from `eslintrc` and replaced `let` with `var` it gives me the expected error.

hzoo:
Can you check eslint 1.10.2 and babel-eslint v5.0.0-beta4?

anshulguleria:
@hzoo  Tried it. Still not working with latest `eslint` and `babel-eslint`

preco21:
I have same issue.
I am using `eslint v2.5.3`, `node v5.9.1`, `babel-eslint v6.0.0`.

danez:
This should be fixed in 6.0.4. Can anyone of you confirm for your usecase?

anshulguleria:
Yup this seems to work fine with `eslint@2.8.0`, `babel-eslint@6.0.4` and `node@6.0.0`.

preco21:
@danez Thanks! It works fine with `eslint@2.10.2`, `babel-eslint@6.0.4` and `node v6.2.0`.

Issue #192, opened 2015-10-07
no-use-before-define doesn't work


mefellows:
The `no-use-before-define` rule doesn't work when the `babel-eslint` parser is used.

Steps to reproduce:

foo.js
```js
console.log(x); // Expect no-use-before-define to throw an error
var x = 1;
```

.eslintrc
```json
{
  "parser": "babel-eslint",
  "rules": {
    "no-use-before-define": 2
  }
}
```
Running 
```
> /tmp/foo $ eslint --config .eslintrc foo.js && echo $?
0
```

Note that removing `"parser": "babel-eslint"` results in the expected behaviour (error is thrown).

Versions:
* eslint version 1.6.0
* babel-eslint version 4.1.3

hzoo:
Ok actually just talked about this in the eslint repo. Workaround for now is to use the `ecmaFeatures` config option https://github.com/eslint/eslint/blob/master/lib/rules/no-use-before-define.js#L98-L104
```js
  "ecmaFeatures": {
    "modules": true
  }
```

mefellows:
Awesome thanks @hzoo, this workaround works nicely for now.

hzoo:
I think this is the best thing to do so closing

Issue #88, opened 2015-05-09
Looking for more maintainers


sebmck:
I don't have time to maintain babel core and babel-eslint so I'm looking for someone(s) to take over maintenance. This would include triaging issues, fixing bugs, reviewing pull requests and releasing new versions. Ideally they'd have experience with AST-based tooling as well as how to write ESLint rules. Familiarity with ESLint internals would be most preferable too so bugs can be identified as residing in the babel-eslint compatibility layer or ESLint itself.

metagen:
To be clear, babel-core is not in need of a maintainer?

sebmck:
Correct, this is only for `babel-eslint`.

hzoo:
I'l be working on supporting jscs with babel with https://github.com/jscs-dev/babel-jscs once I figure out the issues there. If so, some of the issues will probably be the same (esprima/espree) so I maybe of some help although I'm just starting to work with ASTs and eslint.

sebmck:
@hzoo Awesome! Some of the stuff, notably [acorn-to-esprima](https://github.com/babel/babel-eslint/blob/master/acorn-to-esprima.js) could likely be abstracted into a different library (as suggested in #24) that both could use.

bendrucker:
I haven't spent a ton of time with ESLint but I'm willing to step up and take the lead on project management (issues, PRs, releases) and get up to speed on ESLint internals along the way. I'm a big proponent of https://github.com/feross/standard. I'm hoping that if we can get babel-eslint rock solid and demonstrate popular interest I can convince Feross to support custom parsers.

gyandeeps:
@sebmck I would be able to help as I understand eslint.
Do let me know.
Thanks

hzoo:
I think we're good now :smile:

hzoo:
You don't have to be a maintainer, but we're at least looking for some more contributors!

marudor:
I would be willing to contribute as good as I can (I'm probably not familiar enough with eslint/babel to maintain it myself)

hzoo:
Right the issue with this project in particular is that it sort of requires you to have to know about both projects and to fix issues when there might be a minor version change like what happened with https://github.com/babel/babel-eslint/issues/267. Hopefully that doesn't happen again.

vtambourine:
I would love to contribute to this project, since I use it a lot in different places. However I haven't spent much time hacking ESLint core, I am familiar with linters, JSCS and, of course, Babel.

[...]

PR #288, opened 2016-04-01
Fix escope to take sourceType and ecmaVersion from options

danez:
escope was hardcoded to sourcetype: "module" and ecmaVersion: "6"
This changes it to take the configuration from the eslint options and still
defaulting to "module" and "6".
This is done by having the variables in the upper scope, as monkeypatch is only triggered once and would therefore stick to the options that are supplied the first time. But the options can change as in our tests or different eslintrc files in different directories.

Fixes #287 and #217 (217 was failing without the fix in script mode)

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -2,27 +2,32 @@
 "use strict";
 var eslint = require("eslint");
 
-function verifyAndAssertMessages(code, rules, expectedMessages, sourceType) {
-  var messages = eslint.linter.verify(
-    code,
-    {
-      parser: require.resolve(".."),
-      rules: rules,
-      env: {
-        node: true,
-        es6: true
+function verifyAndAssertMessages(code, rules, expectedMessages, sourceType, overrideConfig) {
+  var config = {
+    parser: require.resolve(".."),
+    rules: rules,
+    env: {
+      node: true,
+      es6: true
+    },
+    parserOptions: {
+      ecmaVersion: 6,
+      ecmaFeatures: {
+        jsx: true,
+        experimentalObjectRestSpread: true,
+        globalReturn: true
       },
-      parserOptions: {
-        ecmaVersion: 6,
-        ecmaFeatures: {
-          jsx: true,
-          experimentalObjectRestSpread: true,
-          globalReturn: true
-        },
-        sourceType: sourceType || "module"
-      }
+      sourceType: sourceType
     }
-  );
+  };
+
+  if (overrideConfig) {
+    for (var key in overrideConfig) {
+      config[key] = overrideConfig[key]
+    }
+  }
+
+  var messages = eslint.linter.verify(code, config);
 
   if (messages.length !== expectedMessages.length) {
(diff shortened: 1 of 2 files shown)


josh:
Nice! Thanks @danez

test/non-regression.js
```diff
@@ -1362,6 +1367,54 @@ describe("verify", function () {
...
+      [ "1:5 Implicit global variable, assign as global property instead. no-implicit-globals" ],
+      "script",
+      {
+        env: {},
+        parserOptions: { ecmaVersion: 6, sourceType: "script" }
+      }
+    );
+  });
+
+  it("no-implicit-globals in module", function () {
```
hzoo: I guess we should add a test for when the parserOptions aren't specified?

test/non-regression.js
```diff
@@ -2,27 +2,32 @@
...
-      parserOptions: {
-        ecmaVersion: 6,
-        ecmaFeatures: {
-          jsx: true,
-          experimentalObjectRestSpread: true,
-          globalReturn: true
-        },
-        sourceType: sourceType || "module"
-      }
+      sourceType: sourceType || "module"
```
hzoo: And for that test we may need to remove ` || "module"`?

danez:
I removed the defaulting to `"module"` in the test as @hzoo suggested. Then some tests were failing as the default value for sourceType and ecmaVersion was not working correctly.

After fixing that now only one test fails, but I don't have time right now to investigate. Seems there is still a difference between setting `sourceType='module'` or not setting it.

josh:
> After fixing that now only one test fails, but I don't have time right now to investigate. Seems there is still a difference between setting sourceType='module' or not setting it.

Looks like you're referring to the "no-use-before-define #192" test. It looks like @hzoo suggested modules must be used in this case.

danez:
Ok I finally had time to look at this and figured out what the problem was with the "no-use-before-define" rule test.
The rule was checking the global scope and its first childScope for use-before-define breakages. But babel-eslint (at least in the tests) was setting nodejsScope = true and sourceType module, which created 2 childscopes, one for the nodjes-env and one for the module. And this somehow confused the rule.

Now the nodejsScope is et to false if sourceType is module. Like eslint is doing that: 
https://github.com/eslint/eslint/blob/5685ff58d74cb2b7096bcfa1dfc6cae95c2a9278/lib/eslint.js#L834
https://github.com/eslint/eslint/blob/5685ff58d74cb2b7096bcfa1dfc6cae95c2a9278/lib/eslint.js#L449

josh:
This looks good to me!

danez:
@hzoo Do you have time to look at this?

josh:
Friendly bump 😄

hzoo:
Sorry all - busy and maybe didn't want to check all of this 😄 .

josh:
@hzoo ❤️

hzoo:
Also @josh @danez Do yall want to be contributors? (ref to #88)

PR #295, opened 2016-04-17
Make npm scripts work on Windows too.

fatfisz:
Now they should work everywhere :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,8 @@
     "bootstrap": "git submodule update --init && cd eslint && npm install",
     "eslint": "cd eslint && mocha -c tests/lib/rules/*.js -r ../eslint-tester.js",
     "test": "mocha",
-    "lint": "./node_modules/eslint/bin/eslint.js index.js babylon-to-espree test",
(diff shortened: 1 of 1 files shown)


hzoo:
Thanks

package.json
```diff
@@ -22,8 +22,8 @@
     "bootstrap": "git submodule update --init && cd eslint && npm install",
     "eslint": "cd eslint && mocha -c tests/lib/rules/*.js -r ../eslint-tester.js",
     "test": "mocha",
-    "lint": "./node_modules/eslint/bin/eslint.js index.js babylon-to-espree test",
-    "fix": "./node_modules/eslint/bin/eslint.js index.js babylon-to-espree test --fix",
+    "lint": "node node_modules/eslint/bin/eslint.js index.js babylon-to-espree test",
+    "fix": "node node_modules/eslint/bin/eslint.js index.js babylon-to-espree test --fix",
```
loganfsmyth: Shouldn't these just rely on npm to set the path and be `eslint index.js babylon-to-espree test`?

fatfisz: That would be even better, but then how did the current form come about? 🤔

hzoo: Good point - feel free to send another PR

PR #296, opened 2016-04-17
Fix parent not being set for decorator nodes.

fatfisz:
I was getting strange errors when I put an arrow function into a decorator as an argument (normal function expression was affected as well).
```js
const someObject = {
  @decorator((x) => x)
  method() {}
};
```
The error was coming from the `no-unused-vars` rule; the complaint was this: `Cannot read property 'type' of undefined`, at this line:
```js
if (def.node.parent.type === "Property" && def.node.parent.kind === "set") {
```
So it was obvious for me that the `parent` property of `def.node` was missing.

--
After digging around for quite a long time (I'm unfamiliar with the inner workings of the `eslint` family tools...) I created this solution. Basically this is a monkey patch of `estraverse` that adds visitor key `"decorators"` to appropriate node types. That way decorators will be picked up by the `enter` hook of the `eslint` traverser, and so the `parent` property is set for nodes in the decorator node tree.


hzoo:
It makes sense as the babel ast is somewhat different (in the way that you modified), but it show be covered with `assign(estraverse.VisitorKeys, t.VISITOR_KEYS);`

fatfisz:
So should I make a PR to `babel-types` then?

hzoo:
No because babel doesn't use those ast nodes anymore.

Oh wait I see we're adding decorators 😛... Ok this is correct, we might need to add more types from babel-types based on the ast changes.

fatfisz:
So... yes or no? :smile: And if yes, then where in `babel-types` would it be the best to add these changes?

hzoo:
No I was just saying your PR should be the correct change

fatfisz:
Ok, misunderstood that :)

danez:
Speaking of a more generic solution, we probably should copy the VisitorKeys for all nodes which were renamed in babylon to their espree name before assigning them to the different tools. Is that what you mean @hzoo?

hzoo:
Yeah

fatfisz:
Wouldn't it be ok to patch `babel-eslint` now and refactor later - when there is a need for it? I'm sorry if I'm being ignorant, I just don't understand why it takes so long to ship a fix for a small bug like this :(

hzoo:
Sorry; No reason - just busy and might not be around if we need to revert but we can go ahead do this.

Issue #289, opened 2016-04-04
Comment in SwitchCase triggers newline-before-return


scottohara:
https://github.com/eslint/eslint/issues/5754 describes an issue where a `// no default` comment at the end of a `SwitchStatement` (to satisfy the `default-case` rule when there is no default case) triggers the `newline-before-return`.

This was resolved in `espree` (https://github.com/eslint/espree/pull/257); but a similar fix is required to `babel-eslint`

le0nik:
Same thing with `if` blocks.

Example:

`package.json`

```json
"eslint": "2.8.0",
"babel-eslint": "6.0.3"
```

`.eslintrc.json`

```json
{
    "root": true,
    "parser": "babel-eslint",
    "rules": {
        "newline-before-return": "error",
        "no-console": "error"
    }
}
```

`test.js`

```js
function test() {
    var hasGlobal = Boolean(global);

    if (hasGlobal) {
        /* eslint-disable no-console */
        console.log('test');
        /* eslint-enable no-console */
    }

    return hasGlobal;
}
```

Throws error on `return hasGlobal` line. If you remove `/* eslint-enable no-console */`(the last line in `if` block), the error disappears.

danez:
I created a PR for babylon which fixes this.

zxqfox:
Same with block:
```
a=function() {
  {
    //dummy
  }
  return 15;
}
```

danez:
This was fixed in babylon. Make sure you are using babylon 6.8.2. Or if you don't have babylon in your package.json, just remove node_modules and rerun npm install.

PR #300, opened 2016-04-23
Failing testcase for newline-before-return shouldn't fail with comments.

danez:
Failing testcase for issue #289.
Waiting for babel/babylon#23


danez:
This is now working.

PR #306, opened 2016-04-30
add a message about using the default parser

hzoo:
just wanted to add something like this

Would be useful for most people given I think a lot of folks still think they have to use babel-eslint when they don't have to. ESLint supports ES6 and I still see people thinking it doesn't (same with JSHint even)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@
 **babel-eslint** allows you to lint **ALL** valid Babel code with the fantastic
 [ESLint](https://github.com/eslint/eslint).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,7 +4,9 @@
 **babel-eslint** allows you to lint **ALL** valid Babel code with the fantastic
 [ESLint](https://github.com/eslint/eslint).
 
-`babel-eslint` is successfully linting babel and many other projects.
+## Note: You don't need to use babel-eslint if you are only using ES6/ES2015. ESLint actually supports ES6/ES7, JSX, and object rest/spread by default now.
```
danez: Maybe a little bit smaller, the text is really screaming :)

PR #324, opened 2016-06-17
pin version of babylon to <= 6.8.0 until the problems are fixed

danez:
Should we do that?


hzoo:
Not necessary anymore?

PR #325, opened 2016-06-18
Reset TypeParameters to Identifier

danez:
The new TypeParameter is currently not known by babel-eslint or eslint, so this fixes it by simply making it a Identifier again.

This could probably also have been done by doing more monkeypatching and making escope aware that TypeParameter exists and can be visited.

I also enabled some tests that were skipped, but are now working as it seems.

### DIFF
--- a/babylon-to-espree/toAST.js
+++ b/babylon-to-espree/toAST.js
@@ -167,6 +167,12 @@ var astTransformVisitor = {
       node.type = "Experimental" + node.type;
     }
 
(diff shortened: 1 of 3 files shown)


danez:
@hzoo If you are okay with that we should also create a new release that works completely with babylon 6.8.1

hzoo:
> This could probably also have been done by doing more monkeypatching and making escope aware that TypeParameter exists and can be visited.

This can be a TODO or we move all that functionality out as a separate thing/rule and make babel-eslint just the compat part.

Issue #327, opened 2016-06-20
Expose allowImportExportEverywhere


rhettlivingston:
The Meteor framework recently [recommended ESLint](http://guide.meteor.com/code-style.html#eslint) as the goto lint tool for all of their users. Then, this past week, they added support for imports to be placed below the top-level in conditionals and nested contexts. The change works well and is implemented through the NPM module "reify".

My problem is that whenever either Espree or babel-eslint encounter this, they throw a syntax error that stops all other analysis.

**I am looking for a means that I can specify in the Meteor Guide that all users can use to get around this problem.**

Both Espree and babel-eslint are using parsers that have an option to "allowImportExportEverywhere". I have [created issue #6472 on eslint](eslint/eslint#6472) requesting that they make a change to Espree to pass the option through if specified in .eslintrc.json. They are hesitant to make the change due to their policies against supporting experimental features.

Babel-eslint seems to be the ESLint place for "experimental" changes. But, I don't see how to enable this flag without making a code change in the parseNoPatch routine to set allowImportExportEverywhere to options.allowImportExportEverywhere (when specified).

**Is there a way for me to enable allowImportExportEverywhere without changing code?**

If not, would you consider a PR to add this option to the above code? I'm thinking it would involve adding the line

```js
eslintOptions.allowImportExportEverywhere = options.allowImportExportEverywhere = options.allowImportExportEverywhere || false;
```

to the parse function, changing the existing line that sets allowImportExportEverywhere to false to

```js
allowImportExportEverywhere: options.allowImportExportEverywhere,
```

and test and documentation changes.

hzoo:
We do a similar thing for other options https://github.com/babel/babel-eslint/blob/master/index.js#L366-L367, sounds fine.

benjamn:
When I saw this was closed, I was sure it met with the same YAGNI reply @rhettlivingston got from the ESLint folks—but no! Glad to see this enabled! 🔥 🚒 💃

hzoo:
Heh 😄  in babel-eslint 6.1.0 - and maybe i'l have time later for reify..

PR #329, opened 2016-06-22
Add allowImportExportEverywhere parserOption (#327)

rhettlivingston:
This implements #327 by creating a new parserOption, "allowImportExportEverywhere", that defaults to false and is passed through to babylon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,6 +70,7 @@ Check out the [ESLint docs](http://eslint.org/docs/rules/) for all possible rule
 ### Configuration
 
 `sourceType` can be set to `'module'`(default) or `'script'` if your code isn't using ECMAScript modules.
+`allowImportExportEverywhere` can be set to true to allow import and export declarations to appear anywhere a statement is allowed if your build environment supports that. By default, import and export declarations can only appear at a program's top level.
 
 **.eslintrc**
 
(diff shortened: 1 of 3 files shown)


hzoo:
It would be good if we could write a test for this by passing in the options + code that uses import/export and see that it passes

rhettlivingston:
Agree. Tested it out on my code and ran your existing tests, but should extend yours. Any clues to a starting point or do I need to just dig in?

hzoo:
Yeah basically something like https://github.com/babel/babel-eslint/blob/master/test/non-regression.js#L1412-L1423 where you can pass in some options

rhettlivingston:
Do you think my placement of the option under parserOptions is correct or should I have buried it under parserOptions:ecmaFeatures?

rhettlivingston:
I'm going to answer my own question and move it under ecmaFeatures to match the location of globalReturn since it is very similar in nature.

hzoo:
I think it's fine since this only an option in babylon not eslint anyway?

rhettlivingston:
OK, you sold me. Given my load, that's easy :smile:. So, a test is implemented. Anything else? Other tests?

hzoo:
Sounds good to me!

hzoo:
For babel, we just need to merge something like https://github.com/babel/babel/pull/3503/files to pass babel config options in babelrc to babylon as well

PR #331, opened 2016-06-26
remove hardcoded strict mode value

hzoo:
In babylon

`this.strict = options.strictMode === false ? false : options.sourceType === "module";`

So  can rely on sourceType as the backup

https://github.com/babel/babylon/blob/cde17b33bd6a42fe73289873e3ef62bad2884481/src/tokenizer/state.js#L9


danez:
:+1:

danez:
I always thought we set the correct strictmode option, but that was always only for eslint and not for babylon. :) Good catch.

Issue #307, opened 2016-05-01
"key" property missing from Flow object type aliases


Daniel15:
Example code:

``` js
type MyType = {someProperty: Foo}
```

See http://astexplorer.net/#/n0uGG2VqNE

The `ObjectTypeProperty` should have a `key` property with the `someProperty` identifier and a `value` property with the `GenericTypeAnnotation`. However, the `key` is missing when using babel-eslint on ASTExplorer. Using babylon directly works fine.

hzoo:
That's really odd..

hzoo:
oh sigh lol https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/toAST.js#L175-L178

git blame isn't helping since we moved the file int and out of the repo

hzoo:
Ok https://github.com/babel/babel-eslint/pull/109/files

Daniel15:
So it's actually intentional? Hmm, that's annoying, it makes it impossible to validate Flow object types when using babel-eslint. :disappointed:

hzoo:
Sorry, I think I was just trying to make no-undef and no-unused-vars work for flow and those happened to be exceptions or something but obviously it shouldn't be there. I guess before it was just straight up removing the type altogether. Don't have time to look into this now though

vjeux:
I just ran into a similar issue where the `key` attribute of `ClassProperty` is being removed (example http://astexplorer.net/#/4vTsbwrTFF ).

pranaygp:
I ran into the same issue when writing `eslint-plugin-relay` and found a nice utility used by `eslint-plugin-flowtype`:

https://github.com/gajus/eslint-plugin-flowtype/blob/master/src/utilities/getParameterName.js#L20-L32

Hopefully useful to you if you stumble upon this issue and just want a workaround :)

michaelhogg:
@pranaygp: Thank you, very nice workaround! :tada:

For the benefit of people coming from a search engine and looking for a quick solution, the [code](https://github.com/gajus/eslint-plugin-flowtype/blob/v2.41.0/src/utilities/getParameterName.js#L20-L32) is:

```js
if (node.type === 'ObjectTypeProperty') {

    let tokenIndex = 0;

    if (node.static) {
        tokenIndex++;
    }

    if (node.variance) {
        tokenIndex++;
    }

    const keyIdentifier = context.getSourceCode().getFirstToken(node, tokenIndex);

    // keyIdentifier has this structure:
    // {
[...]

PR #335, opened 2016-07-08
Rename ClassProperty.key instead of deleting it

vjeux:
The no-undef rule uses escope which blindly looks for a key attribute in order to know if something is in scope. Right now the fix is to delete the key attribute but I need it for a rule I'm writing at Facebook. Instead of deleting it, just renaming it to _key works.

For context, the rule I'm writing is checking if a bound function using property initializer is not needed. We're going to use this for our React.createClass to es6 class migration in Facebook codebase.

Test Plan:
All the tests run fine.

### DIFF
--- a/babylon-to-espree/toAST.js
+++ b/babylon-to-espree/toAST.js
@@ -208,6 +208,7 @@ var astTransformVisitor = {
 
     // remove class property keys (or patch in escope)
     if (path.isClassProperty()) {
+      node._key = node.key;
       delete node.key;
     }
 


vjeux:
Fixes one of the issues in https://github.com/babel/babel-eslint/issues/307

hzoo:
Yeah sounds good. Could also figure out another way but this is a good workaround. We did a similar thing for the type as well with ` node._babelType = node.type;`

Maybe generalizing that for other things that were deleted with a prefix would be useful as well

vjeux:
Happy to rename it to _babelKey and/or making it a bit more generic to handle all the things we delete in this file. Let me know what you want me to do :)

hzoo:
Do you think `_babelKey` is more clear? I think either is ok. Also not sure why its only failing on < node 4

vjeux:
I honestly don't care how it is named as long as I can read it :)

vjeux:
Let me look at the errors

hzoo:
I was at some point going to check if https://github.com/zertosh/eslint-plugin-flow-vars covered class properties since then we could drop all the escope stuff for flow vars and not have to change the AST as much

vjeux:
Hmm weird, they are passing on my laptop with node 4. I force pushed to re-run travis

vjeux:
cc @zertosh :)

hzoo:
I think we can have another PR to add the other ones. like `._async` or something or some other way to specify things that had to be removed. Although maybe that should be discussed since we are modifying the AST and we'll need to keep it in for compatibility moving forward

vjeux:
I can repro the errors on node 0.10, looking

hzoo:
I checked out the pr locally and looks like the tests pass for me on 0.10

vjeux:
Did you also regenerate all the node_modules?

hzoo:
I see it fails when `npm install` on node 0.10 so it sounds like something weird with npm 2 to me

vjeux:
Grrr, so the scope thing actually doesn't care about the name of the attribute, it checks for all the nodes of type "Identifier" and checks for node.parent.type but there's no parent in the babylon node so it fatals.

If I do

```
node._getKey = function() { return node.key; }
```

then it doesn't crash...

vjeux:
Let me know what you suggest I should do :)

hzoo:
Maybe we could mark all ClassProperty as defined and not delete the key

vjeux:
Sounds like a plan, is there some existing code that mark nodes as being defined I can base on, I'm very new to this whole codebase.

hzoo:
Yeah for sure (wasn't expecting you to do it 😄)

vjeux:
I'm ready to do whatever is needed to get this shipped :) Thanks a lot for all your help so far, I really appreciate

hzoo:
Now that I think about it, most of the code does the `visiting` to mark it as used rather than mark it as defined.

I think we would basically to the same thing as the flow plugin https://github.com/zertosh/eslint-plugin-flow-vars/blob/master/define-flow-type.js

Ah I remember we did it for https://github.com/babel/babel-eslint/blob/master/index.js#L319-L330

hzoo:
Or I could join reactiflux

I think I got it

```js
// in visitClass
var classBody = node.body.body;
for (var a = 0; a < classBody.length; a++) {
      if (classBody[a].type === "ClassProperty") {
        createScopeVariable.call(this, classBody[a], classBody[a].key);
      }
    }
```

vjeux:
Okay, looks like you are making much more progress than I am :p

hzoo:
Should probably write another test but https://github.com/babel/babel-eslint/pull/336/files

hzoo:
Either way there's a similar issue with the other ones


```
    // async function as generator
    if (path.isFunction()) {
      if (node.async) node.generator = true;
    }
```

super awkward since it's obviously not a generator

thejameskyle:
Maybe we can simplify a lot of this by cloning the original babel Node an having an accessor for it.

hzoo:
Yeah possibly; For async/await I was thinking maybe we don't transform them and just create a new rule for the broken rules in eslint-plugin-babel.

PR #352, opened 2016-07-29
Fix tests for eslint 3.2

danez:
The added a dot at the end of the no undef message. Now the tests fail either for version 2 or 3 :(

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -402,7 +402,7 @@ describe("verify", function () {
           "var b: T = 1; b;"
         ].join("\n"),
         { "no-unused-vars": 1, "no-undef": 1 },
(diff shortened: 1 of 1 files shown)


jmeas:
> Now the tests fail either for version 2 or 3 :(

lol 😓

hzoo:
sigh, we should just drop support for eslint 2 at this point then?

jmeas:
Seems reasonable to me @hzoo.

hzoo:
Since we don't use peerDeps, if we are dropping we should a major for that?

jmeas:
> Since we don't use peerDeps, if we are dropping we should a major for that?

Being conservative about it seems good. I'm :+1: to a major bump.

Issue #323, opened 2016-06-13
Dropping Node < 4


hzoo:
Given eslint 3.0 will drop it, I guess we should as well as another major version?

https://github.com/eslint/eslint/issues/4483, https://github.com/eslint/eslint/pull/6401

kaicataldo:
Does this involve any code changes on this side? Asking because the ESLint team is talking about starting to use some Node 4 features that we weren't using before.

danez:
I think it will depend on what else changes in eslint 3. If the same version of babel-eslint will work for eslint 2.x and 3.x then we might not do immediately. On the other hand if we have to change some stuff anyway to support (only) 3.x then maybe yes.

hzoo:
Ok we I think we can drop it now since the tests are currently failing on either eslint 2 or eslint 3

hzoo:
Should do a major/drop for https://github.com/babel/eslint-plugin-babel too

PR #358, opened 2016-08-01
Drop node < 4

hzoo:
Fixes https://github.com/babel/babel-eslint/issues/323

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,9 @@
   },
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "license": "MIT",
+  "engines": {
+    "node": ">=4"
+  },
   "bugs": {
     "url": "https://github.com/babel/babel-eslint/issues"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,6 +29,9 @@
   },
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "license": "MIT",
+  "engines": {
+    "node": ">=4"
```
jmeas: Should there be an upper limit placed here? Although it's probably pretty unlikely that a future Node would break this library, we can't know for sure until they're released :)

jmeas:
lgtm 👍

package.json
danez: If a new node version breaks, we have to take action in any case. If it not breaks, then in this case we wouldn't have to do anything, but with limit we would always need to care. So I think it is fine like it is :)

Issue #357, opened 2016-08-01
babel-eslint 7.x


hzoo:
### ESLint Compat
- [x] drop node < 4 like ESLint #323
- [ ] #350 just do it now actually since it's a breaking change to change the AST output
  - [ ] Look for other places in https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/toAST.js that we can remove to have a similar AST
- [ ] https://github.com/babel/eslint-plugin-babel drop node 4 as well

### More stable library
- [ ] Less code: lets remove the flow code for unused-vars and just recommend ~~https://github.com/zertosh/eslint-plugin-flow-vars~~  [eslint-plugin-flowtype](https://github.com/gajus/eslint-plugin-flowtype/issues/35). Maybe we can investigate automatically including it (but it would probably involve monkey patching which we don't want) - remove almost everything in https://github.com/babel/babel-eslint/blob/master/index.js#L128-L213
  - [ ] Bring back marking nodes as used as a seperate plugin if users want it
- [ ] figure out how to remove the scope adding/etc and do it another way

### Misc
- [ ] maybe figure out better error messages?

danez:
👍 
We should setup the current tests to include eslint-plugin-flow-vars, so that we still se what is unsupported even when using the plugin.

niksajanjic:
Hi, maybe this might not be the perfect place to mention this, but you guys are free to create separate issue if needed. At the moment if one would like to install this module using npm you would get deprecation warning:

npm WARN deprecated lodash.assign@4.2.0: This package is deprecated. Use Object.assign.

I'm guessing this could be solved with next version. Our team gets this warning using node 5 or 6 and npm version 3.

hzoo:
Ok went a little crazy with what we should do, removing flow code etc since the other flow plugin rules depend on it actually. Just going to drop node 4, fix async/await so we can get a release out #361

Issue #350, opened 2016-07-29
Remove the ast changes for async/await when ESLint supports it


hzoo:
https://github.com/babel/babel-eslint/blob/aff95b8a5a0853f6bd8cb9b19cd14d3c202c3dd9/babylon-to-espree/toAST.js#L209-L226

```
    if (path.isFunction()) {
      if (node.async) node.generator = true;
    }

    // await transform to yield
    if (path.isAwaitExpression()) {
      node.type = "YieldExpression";
      node.delegate = node.all;
      delete node.all;
    }
```

Would need to modify the generator star spacing rule so we can also remove it in eslint-plugin-babel

tomkel:
eslint 3.6.0 supports es2017 now

hzoo:
Doing this in https://github.com/babel/babel-eslint/issues/361

PR #361, opened 2016-08-03
remove eslint 2 logic

hzoo:
ref https://github.com/babel/babel-eslint/issues/357

- [x] remove estraverse-fb logic
- [x] remove logic for accounting for async/await before eslint supported it https://github.com/babel/babel-eslint/issues/350

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -11,7 +11,7 @@ function verifyAndAssertMessages(code, rules, expectedMessages, sourceType, over
       es6: true
     },
     parserOptions: {
-      ecmaVersion: 6,
+      ecmaVersion: 8,
       ecmaFeatures: {
         jsx: true,
         experimentalObjectRestSpread: true,
--- a/babylon-to-espree/toAST.js
+++ b/babylon-to-espree/toAST.js
@@ -202,23 +202,11 @@ var astTransformVisitor = {
       }
     }
 
-    // async function as generator
-    if (path.isFunction()) {
-      if (node.async) node.generator = true;
-    }
-
     // TODO: remove (old esprima)
     if (path.isFunction()) {
       if (!node.defaults) {
         node.defaults = [];
       }
-    }
-
-    // await transform to yield
-    if (path.isAwaitExpression()) {
-      node.type = "YieldExpression";
-      node.delegate = node.all;
-      delete node.all;
     }
 
     // template string range fixes
--- a/index.js
+++ b/index.js
@@ -1,11 +1,11 @@
 var babylonToEspree = require("./babylon-to-espree");
-var pick           = require("lodash.pickby");
-var Module         = require("module");
-var path           = require("path");
-var parse          = require("babylon").parse;
-var t              = require("babel-types");
-var tt             = require("babylon").tokTypes;
-var traverse       = require("babel-traverse").default;
+var pick            = require("lodash.pickby");
+var Module          = require("module");
+var path            = require("path");
+var parse           = require("babylon").parse;
+var t               = require("babel-types");
+var tt              = require("babylon").tokTypes;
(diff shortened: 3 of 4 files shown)


danez:
How do you want to fix the tests? By adding eslint-plugin-flow?

hzoo:
Yeah and we would recommend doing that somehow (would be nice if we could warn on error to do that)

danez:
Ok I'm going to try to use flow-plugin.

danez:
Seems that eslints node API does not support plugins  😥
https://github.com/eslint/eslint/issues/6841

hzoo:
Yeah although that's a different issue. The parser can't include any plugins - I just meant for our tests?

danez:
That is what I'm trying only for the tests of course, I finally got it working by using `eslint.CLIEngine` which feels a little bit wrong because we are not in an cli but ¯\_(ツ)_/¯

But with the plugin enabled nothing changed and there are still the same errors. I think I commit what I have, although it does not change anything, but maybe you see whats wrong.

danez:
I think the plugin is relying on the fact that we visit also type annotations, which we now don't do anymore. https://github.com/zertosh/eslint-plugin-flow-vars/blob/master/use-flow-type.js#L12

hzoo:
Oh interesting! @zertosh

danez:
I found the problem, it was the removal of the monkeypatch for the "other" estraverse. This made the failure count go from 60 to 10.

But not sure if we really need to patch both estraverses :D

hzoo:
yeah wasn't sure if we needed to do it anymore - https://github.com/eslint/eslint/issues/6364, https://github.com/eslint/eslint/pull/5543

danez:
Seems that the changes to estraverse (which allow fallback to be a function) were only made to support unknown node types. So in case of nodes that are known by [estraverse](https://github.com/estools/estraverse/blob/master/estraverse.js#L218) you need to overwrite the internal VisitorKeys, as the `fallback` will only be used when there is [no entry in the VisitorKeys](https://github.com/estools/estraverse/blob/master/estraverse.js#L517). You can overwrite the keys in the estraverse API but only if you are the one calling `traverse()`.

In our case I don't see any other way currently besides monkeypatching. The only other solution would be if eslint would allow the parser to supply custom VisitorKeys.

But maybe I'm also missing something. /cc @nzakas

hzoo:
I remember I added it but it was removed in https://github.com/babel/babel-eslint/pull/273 @mysticatea

zertosh:
>> I think the plugin is relying on the fact that we visit also type annotations

> Oh interesting! @zertosh is that right?

Yup.

> Also we'll want to use https://github.com/gajus/eslint-plugin-flowtype later since ^ will be deprecated

Also, yup. I'm going to up put the notice now. It made sense to merge `eslint-plugin-flow-vars` into `eslint-plugin-flowtype`.

danez:
I just linked this branch of babel-eslint to eslint-plugin-flowtype and 31 tests then fail there. This is due to the fact that eslint-plugin-flowtype assumes that some things are declared or visited. So the plugin needs to be enhanced to cover all this cases, but it seems pretty easy to add.

hzoo:
Ok maybe we just need the logic in for now and just drop the extra monkey patching above + node < 4 and work doing the other stuff in another pr

test/non-regression.js
```diff
@@ -11,7 +11,7 @@ function verifyAndAssertMessages(code, rules, expectedMessages, sourceType, over
       es6: true
     },
     parserOptions: {
-      ecmaVersion: 6,
+      ecmaVersion: 8,
```
danez: maybe lets use years here? 2017

tomkel: I agree, years please

PR #410, opened 2016-10-11
Now using template strings

stevenjob:
The problem with this PR is that using multiline template literals causes indentation to show in the new multiline string where it did not previously with the array join. (an extra 6 spaces is now before most lines)

I have used template literals in all places and the tests all still pass.

### DIFF
--- a/babylon-to-espree/toAST.js
+++ b/babylon-to-espree/toAST.js
@@ -160,7 +160,7 @@ var astTransformVisitor = {
     }
 
     if (path.isRestProperty() || path.isSpreadProperty()) {
-      node.type = "Experimental" + node.type;
+      node.type = `Experimental${node.type}`;
     }
 
     if (path.isTypeParameter && path.isTypeParameter()) {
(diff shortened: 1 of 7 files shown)


hzoo:
Awesome

hzoo:
We can use something we do in babili - unpad function 
https://github.com/babel/babili/blob/master/utils/unpad.js

usage: https://github.com/babel/babili/blob/master/packages/babel-plugin-transform-property-literals/__tests__/transform-property-literals-test.js#L27

stevenjob:
Ok will do this, thanks

stevenjob:
Unpad should probably be its own module but i have just copied it to a utils folder for now.

I have used unpad in each test separately as it was simple to add with regex and i did not want to modify parseAndAssertSame or  verifyAndAssertMessages. I assumed there could be use cases where you would want leading spaces in some test code.

hzoo:
Yeah looks like there is a module for it that was from babili/babel originally https://www.npmjs.com/package/unpad but this is fine

PR #413, opened 2016-10-17
[import()] Adding support to lint dynamic imports

kesne:
This adds the `dynamicImport` to the plugin list, which allows `import()` to be parsed.

@hzoo I'm working on getting a babel plugin to transpile these calls into something webpack 1 understands. Would be great to get linter support landed!


kesne:
Added a simple test to prevent regression.

PR #418, opened 2016-10-28
append code frame on parse error

hzoo:
After seeing https://github.com/eslint/eslint/pull/7437,

I noticed that parser errors didn't have a codeframe either so figured we could add it there too? Just an idea and does change the error messages of course then

screen shot 2016-10-28 at 5 33 11 pm

screen shot 2016-10-28 at 5 32 58 pm
- [ ] fix the tests


danez:
👍

Issue #397, opened 2016-09-27
Using Node 4 Features


hzoo:
Since we dropped node < 4,

an easy win would be to change our tests to use template strings instead of

``` js
["asdfads",
"asdf"
].join('\n')
```

This doesn't have to be done in a single PR, and would be good for new contributors

vgvenkat:
Hello there! Would this be conversion of all strings which use string concat using "+" in the [3 js test files](https://github.com/babel/babel-eslint/tree/master/test) as well or just the ones with ```join``` to use template strings?

hzoo:
@vgvenkat Sorry there's already a pr for that particular change https://github.com/babel/babel-eslint/pull/410 (should of put a notice)

vgvenkat:
oh. ok cool 👍🏻

kaicataldo:
We did the same thing in the ESLint repo recently (there's one PR open that adds the final thing we wanted). This list might be helpful: https://github.com/eslint/eslint/issues/6407

kaicataldo:
ESLint itself also has some fixers that could be used - `prefer-template` was just added recently and I used it on the main ESLint repo with no issues!

karanjthakkar:
@hzoo Since #410 is merged, is there any other change that this issue is looking to address? I'd love to help out with other changes, if any, apart from the conversion to template strings.

kaicataldo:
There actually is a PR open with some of this work done: https://github.com/babel/babel-eslint/pull/425

hzoo:
Going to close as some of this is done and don't need to track

PR #425, opened 2016-11-18
Update to use Node 4 features

nhajidin:
Some changes for #397

Mostly replacing `for` with `forEach` and using arrow functions.


kaicataldo:
Hey, thanks for contributing! I'm not 100% behind the switch from `for` loops to `forEach()`, since it seems like an unnecessary change, but I think everything else looks great.

nhajidin:
I've reverted all the forEach changes.

hzoo:
Sorry for the delay, lgtm! Thanks @nhajidin

PR #435, opened 2017-01-28
Use `lodash` instead of `lodash.pickby`.

wtgtybhertgeghgtwtg:
Since `babel-types` and `babel-traverse` require `lodash`, it's already in the dependency tree.  `lodash.pickby` is not.
So, ironically, using `lodash.pickby` increases the dependencies, instead of decreasing them.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 var babylonToEspree = require("./babylon-to-espree");
-var pick            = require("lodash.pickby");
+var pick            = require("lodash").pickBy;
 var Module          = require("module");
 var path            = require("path");
 var parse           = require("babylon").parse;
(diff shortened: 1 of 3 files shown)


xtuc:
We could even replace `pickby` with the native filter. Thus removing the dependency.

wtgtybhertgeghgtwtg:
So
```
t.VISITOR_KEYS.filter
```
instead of
```
pick(t.VISITOR_KEYS,
```
I'm ashamed I didn't think of that.

xtuc:
Yes, no worries :+1:. We appreciate your contribution.

Some related PR: babel/babel#5042, babel/babel#5043, babel/babel#5180 and babel/babel#5181

wtgtybhertgeghgtwtg:
Actually, that wouldn't work, since `t.VISITOR_KEYS` is an object used as a map, not an array.

xtuc:
Ok, I was confused by the naming. `pick` should be fine.

hzoo:
Thanks again @wtgtybhertgeghgtwtg!!

PR #446, opened 2017-03-19
Add option to disable code frame.

Couto:
This might sound weird, since code frame is a great feature of babel-eslint.

My use case:

I want to display eslint reports in [go.cd](https://www.gocd.io/), for that I need to generate an HTML report, which is usually easy with eslint HTML formatter, and it results in something like this:

screen shot 2017-03-19 at 14 14 07

However with babel-eslint, due to the code frame feature the same report looks like this:

screen shot 2017-03-19 at 14 15 45

This PR adds an option to `parserOptions` to disable the code highlight feature, and when enabled, it simply returns the regular exepected output to eslint, so... when `.eslintrc` looks something like this:

```
{
  "parser": "babel-eslint",
  "parserOptions": {
    "codeFrame": false
  },
  "extends": "eslint:recommended"
}
```

the generated HTML will look similar to this:

screen shot 2017-03-19 at 14 20 59

I hope this PR is good enough to merge, however if anything is needed, please just say so :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -366,6 +366,7 @@ exports.parse = function (code, options) {
 
 exports.parseNoPatch = function (code, options) {
   var opts = {
+    codeHighlight: options.hasOwnProperty("codeHighlight") ? options.codeHighlight : true,
     sourceType: options.sourceType,
     allowImportExportEverywhere: options.allowImportExportEverywhere, // consistent with espree
     allowReturnOutsideFunction: true,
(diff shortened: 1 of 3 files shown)


hzoo:
Sounds good, cc @kaicataldo on where to put this option, sounds like it has to be in parserOptions

kaicataldo:
I think keeping it in `parserOptions` is good

hzoo:
Can we add a test somewhere for this - that it prints/doesn't print those characters? Maybe can just write a syntax error and regex/search for a char?

Couto:
I've just added some basic tests as suggested. However I'm not sure of their correct place inside the test files. If they are not in right place, please just say so. :)

hzoo:
maybe it could go in it's own file but this is good!

hzoo:
Thanks @Couto, and can you add this to the readme?

hzoo:

PR #447, opened 2017-03-20
Chore: Clean up and upgrades

kaicataldo:
This is cleanup/upgrading deps in preparation for trying to use Babylon with the ESTree plugin. The current ESLint tests simply do not work. I think we need to remove the current broken code and rethink this after we start using the ESTree plugin.


hzoo:
Awesome!

Issue #445, opened 2017-03-18
[flow] declare function does not recognize bounded polymorphic types


rattrayalex:
In the course of https://github.com/babel/babel-eslint/pull/444 , I noticed that

```js
    it("declared function type parameter bounds", () => {
      verifyAndAssertMessages(
        unpad(`
          import type Foo from 'foo';
          declare function foo<T: Foo>(a: T): T;
          foo(1);
        `),
        { "no-unused-vars": 1, "no-undef": 1 },
        []
      );
    });
```
results in 
```json
[
   {
      "ruleId":"no-unused-vars",
      "severity":1,
      "message":"'Foo' is defined but never used.",
      "line":1,
      "column":13,
      "nodeType":"Identifier",
      "source":"import type Foo from 'foo';"
   }
]
```
and
```js
    it("declared class type parameter bounds", () => {
      verifyAndAssertMessages(
        unpad(`
          import type Foo from 'foo';
          declare class C<T: Foo> {
            x: T
          };
          new C();
        `),
        { "no-unused-vars": 1, "no-undef": 1 },
        []
      );
    });
```
results in 
```json
[
   {
      "ruleId":"no-unused-vars",
      "severity":1,
      "message":"'Foo' is defined but never used.",
      "line":1,
      "column":13,
      "nodeType":"Identifier",
      "source":"import type Foo from 'foo';"
   },
   {
      "ruleId":"no-unused-vars",
      "severity":1,
      "message":"'T' is defined but never used.",
      "line":2,
      "column":17,
      "nodeType":"Identifier",
      "source":"declare class C<T: Foo> {"
   }
]
```

I don't use Flow much, so I'm not encountering this in the wild, but trying the above at https://flowtype.org/try/ appears to indicate that the code is semantically correct.

zertosh:
@rattrayalex #459 now fixes this, but I need to split up that PR

rattrayalex:
Nice!

pronebird:
I have similar case:

```
declare class ClipboardData {
  setData(type: string, data: string): void;
  getData(type: string): string;
}

declare class ClipboardEvent extends Event {
  clipboardData: ClipboardData;
}
```

Output:

>'ClipboardData' is defined but never used.

kaicataldo:
[...]

PR #449, opened 2017-03-22
Fix type param and interface declaration scoping

zertosh:
In #444, @rattrayalex, fixed type param scoping for functions but didn't cover classes, type aliases or interfaces - this PR fixes that. Also this adds scoping for interfaces since you can't scope type params in interfaces without first scoping interfaces.

As pointed out in #445, type params in declarations is broken. While preparing this PR, I noticed that pretty much all `declare` anything are broken. I'll follow up with a PR for that.

### DIFF
--- a/index.js
+++ b/index.js
@@ -176,18 +176,6 @@ function monkeypatch() {
     }
   }
 
-  function visitTypeParameters(typeParameters) {
-    var params = typeParameters.params;
-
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -176,18 +176,6 @@ function monkeypatch() {
     }
   }
 
-  function visitTypeParameters(typeParameters) {
```
rattrayalex: nice, your approach looks much cleaner

PR #450, opened 2017-03-22
Remove lodash dependency

zertosh:
#435 switched `lodash.pickby` to `lodash` to reduce the size of the dependency tree (since `lodash` would be shared with `babel-types` and `babel-traverse`). Although it did do that, it ended up loading all of `lodash`. It should've done `require('lodash/pickBy')` instead of `require('lodash').pickBy`. However, since `babel-*` doesn't use `pickBy`, we'd load those deps for no reason. Since it's a trivial function to implement with native functions, this PR does that instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,4 @@
 var babylonToEspree = require("./babylon-to-espree");
-var pick            = require("lodash").pickBy;
 var Module          = require("module");
 var path            = require("path");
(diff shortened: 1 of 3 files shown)


hzoo:
True, good idea!

hzoo:

zertosh:
@hzoo :+1: thanks!

Issue #440, opened 2017-02-23
Use Babylon ESTree plugin


hzoo:
to remove a lot of the ast hacks that are already done via the plugin.

https://github.com/babel/babylon#output

basically everything in https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/toAST.js

This will make this project a lot simpler and my soul ☺️

alxpy:
Could I work on it?

kaicataldo:
I'd like to work on this!

kaicataldo:
Haven't had the time, unfortunately - if anyone else is interested, please feel free!

StGeass:
Well, then i'd like to take it

PR #455, opened 2017-03-25
Tidy up babylon-to-espree

zertosh:
The only "breaking" change are the exports of `babylon-to-espree`. That wasn't part of the public API, so hopefully nobody is using them.

### DIFF
--- a/babylon-to-espree/attachComments.js
+++ b/babylon-to-espree/attachComments.js
@@ -1,3 +1,5 @@
+"use strict";
+
 // comment fixes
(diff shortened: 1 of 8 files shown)


hzoo:
btw another thing we should do is https://github.com/babel/babel-eslint/issues/440 which @kaicataldo wanted to work on - should remove a lot of this stuff hopefully

zertosh:
@hzoo I just tried babylon@7 with the estree plugin (still needs tokens, move Program<->File, and few more things) and the tests mostly pass with master :)

PR #458, opened 2017-03-26
RFC: Do not set a default ecmaVersion or sourceType

zertosh:
This is a super breaking change, but it ensures correctness in ESLint's rules, and eliminates surprises.

Before this change, `babel-eslint` set defaults for `sourceType` and `ecmaVersion` that it applied to Babylon and escope. However, these values could not propagate to the ESLint rules themselves or ESLint APIs. This resulted in incorrect behavior by certain rules. Consider these examples:

* https://github.com/eslint/eslint/blob/6a718ba/lib/eslint.js#L1093-L1101
* https://github.com/eslint/eslint/blob/6a718ba/lib/rules/func-name-matching.js#L40-L51
* https://github.com/eslint/eslint/blob/6a718ba/lib/rules/no-redeclare.js#L86-L98

Making things even more surprising is the fact that setting `env: es6` sets the `ecmaVersion` to `6` but doesn't change `sourceType` (https://github.com/eslint/eslint/blob/6a718ba/conf/environments.js#L98-L103).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,10 @@ $ npm install eslint@3.x babel-eslint@6 --save-dev
 ```json
 {
   "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaVersion": 8,
+    "sourceType": "module"
+  },
   "rules": {
     "strict": 0
   }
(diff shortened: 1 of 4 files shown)


hzoo:
I guess we should do this to keep consistent behavior in ESLint, would be breaking but we are doing a v7 atm.

README.md
```diff
@@ -76,19 +80,42 @@ Check out the [ESLint docs](http://eslint.org/docs/rules/) for all possible rule
 
 ### Configuration
 
-`sourceType` can be set to `'module'`(default) or `'script'` if your code isn't using ECMAScript modules.
-`allowImportExportEverywhere` can be set to true to allow import and export declarations to appear anywhere a statement is allowed if your build environment supports that. By default, import and export declarations can only appear at a program's top level.
-`codeFrame` can be set to false to disable the code frame in the reporter. This is useful since some eslint formatters don't play well with it.
+* `sourceType`, `ecmaVersion`, `ecmaFeatures.globalReturn` and `ecmaFeatures.impliedStrict` should be set in accordance with http://eslint.org/docs/user-guide/configuring#specifying-parser-options.
+  - `ecmaVersion` is disregarded by `babel-eslint`, however, some ESlint rules depend on this value being set correctly.
```
ljharb: What would it take to make the first part not true?

README.md
```diff
@@ -76,19 +80,42 @@ Check out the [ESLint docs](http://eslint.org/docs/rules/) for all possible rule
...
+  - If `sourceType` is not set to `"module"`, then using `import`/`export` will error.
```
ljharb: Also, files will not automatically be in strict mode.

hzoo:
I guess the reason we had the defaults earlier though is that babylon automatically parses what is spec so the ecmaversion should at least be es2017 atm? and we assume modules normally

kaicataldo:
Unless I'm missing something, seems to me we shouldn't remove the defaults. I agree with @hzoo - I think default for `ecmaVersion` should be the latest version available in ESLint and we should leave the default for "sourceType" as well. So to me it seems like the changes we might want are to update the `ecmaVersion` default to `2017` and then add documentation saying that the config should also include the following to make sure that ESLint core is doing the right thing:
```json
"parserOptions": {
    "ecmaVersion": 8
},
```

ljharb:
Please let's make sure documentation uses year, not edition number - it was ES6/ES2015, but nobody calls it ES8 - it's ES2017.

danez:
Related to this: I created this ticket eslint/eslint#8744 which is the opposite to this PR. :)
Basically what I thought would be nice is if the parser could decide what the default options are in a way that eslint understands it and uses them everywhere correctly.

zertosh:
@kaicataldo, I'd love to, but I just don't have time these days :(

kaicataldo:
Okay! Will close this, but happy to have this discussion again in the future!

Issue #438, opened 2017-02-12
New https://github.com/eslint/eslint-scope project


JamesHenry:
Hey @hzoo et al 😄 

I just wanted to make sure you are aware that we have officially forked escope at https://github.com/eslint/eslint-scope

There were a couple of reasons for this, but most recently it was not possible to work with the maintainers of escope to accept PRs relating to extensions to ESTree (e.g. for features relating to typeAnnoations)

We will be working to replace escope with eslint-scope in ESLint as part of the next major release (v4).

@nzakas was keen that you guys were aware of that in plenty of time, as I know you currently have to do some funky stuff around scopes in this project.

Please let us know if you have any questions!

zertosh:
@JamesHenry, I'm actively working on rewriting this package so that flow types are perfectly scoped. Not only so that eslint can make better use of the scope info, but also to power other fb tooling that needs type data (e.g.: code generators that need to resolve types). Would you accept a series of PRs in `eslint-scope` that add type annotation support?

soda0289:
@zertosh I have been working on adding type annotation and decorator support to eslint-scope, since typescript would benefit from this as well. If we added type parameter support as well would that cover everything for flow?

hzoo:
@vitorbal updated to eslint-scope in https://github.com/babel/babel-eslint/pull/461 but we should at least update the devDep to check that tests all still pass https://github.com/babel/babel-eslint/blob/1aedb95ef40c5d51223831dfd8be65efbdc27e09/package.json#L40

hzoo:
I think we can close now since we use this as a dep
https://github.com/babel/babel-eslint/blob/bf27f6021e3a3661589bfa1e314f45876edeb20b/package.json#L18

PR #459, opened 2017-03-26
[WIP] Refactor visiting strategy

zertosh:
This PR does away with the custom traverser, and instead adds correct node visitors to get accurate scopes (and references/definitions) for flow types. It's written for easy upstreaming into `eslint-scope` (see #438) if that becomes a thing. How? New node type visitors are added (look for `//new`), and only node type visitors were overwritten (look for `//override`) instead of `visit*` functions. There's some copy/paste from escope in the overwritten methods to clearly show where the new flow type code must go (look for `<custom></custom>`).

#### TODO:

- [ ]  Fix `declare` (Correct scoping, especially with `declare module`).
- [ ] More scope tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,5 @@
+"use strict";
+
 var babylonToEspree = require("./babylon-to-espree");
 var Module          = require("module");
 var path            = require("path");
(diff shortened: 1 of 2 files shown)


hzoo:
fyi eslint pr is https://github.com/eslint/eslint/pull/8280

kaicataldo:
Is this still something we want to do?

hzoo:
and @zertosh if you have time to work on this again, I think a refactor PR sounds great!

PR #478, opened 2017-05-28
Fix: Convert RegExpLieteral value to RegExp object (fixes #477)

### LINKED ISSUE #477: RegExpLiteral node needs to have RegExp Object value
Currently babel-eslint converts the `RegExpLiteral` to a estree `Literal` with a value of `{}`.
https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/toAST.js#L97

ESLint expects the value to be a RegExp object. This allows it to be casted to a string correctly. It should be `null` when the flags u or y are unsupported.
https://github.com/estree/estree/blob/master/es5.md#regexpliteral

I worked on a fix recently for `typescript-eslint-parser` and can make a fix for this parser as well. It causes the rule `no-dupe-keys` to fail.
https://github.com/eslint/eslint/issues/8655

soda0289:
The only way I could get the tests to fail was to check the constructor name of both objects. I tried to match them for every object but many times they do not match.

### DIFF
--- a/babylon-to-espree/toAST.js
+++ b/babylon-to-espree/toAST.js
@@ -94,7 +94,11 @@ var astTransformVisitor = {
     if (path.isRegExpLiteral()) {
       node.type = "Literal";
       node.raw = node.extra.raw;
-      node.value = {};
+      try {
+        node.value = new RegExp(node.pattern, node.flags);
+      } catch (err) {
+        node.value = null;
+      }
       node.regex = {
         pattern: node.pattern,
         flags: node.flags
--- a/test/babel-eslint.js
+++ b/test/babel-eslint.js
@@ -20,6 +20,8 @@ function assertImplementsAST(target, source, path) {
   var typeB = source === null ? "null" : typeof source;
   if (typeA !== typeB) {
     error(`have different types (${typeA} !== ${typeB}) (${target} !== ${source})`);
+  } else if (typeA === "object" && ["RegExp"].indexOf(target.constructor.name) !== -1 && target.constructor.name !== source.constructor.name) {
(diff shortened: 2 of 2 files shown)


babylon-to-espree/toAST.js
```diff
@@ -94,7 +94,11 @@ var astTransformVisitor = {
     if (path.isRegExpLiteral()) {
       node.type = "Literal";
       node.raw = node.extra.raw;
-      node.value = {};
+      try {
+        node.value = new RegExp(node.pattern, node.flags);
+      } catch (err) {
+        node.value = null;
```
kentcdodds: Could you give a scenario in tests that handles when the `value` is `null`? Also, perhaps we should have it default to `{}` as it was before? Otherwise this could be a breaking change right?

soda0289: A scenario where is is null is when the flags `u` and `y` are not supported. I have tests cases for these and it will produce null when run in node v4 and v5. 

I'm not sure this is a breaking change since the output is was producing is a bug and did not match what espree, or esprima produce. The reason the test pass is it only compared the `typeof`, which were both objects, but not from the same class. We could set the default to `{}` but that would not match the ESTree spec for RegExp Literal nodes.
https://github.com/estree/estree/blob/master/es5.md#regexpliteral

This does fix an issue with eslint https://github.com/eslint/eslint/issues/8655 and setting the default to `{}` might cause other rule failures.

The estree plugin for babylon does produce the correct output so if this were to be considered a breaking change we could wait till that is implemented in this parser.

kentcdodds: You've convinced me. Thanks 😄

PR #486, opened 2017-06-11
updated package.json

EsrefDurna:
updated package eslint from 3.18.0 to 3.19.0


hzoo:
`^` covers that change so we don't need to update. thanks

PR #489, opened 2017-06-15
Use babylon estree and ranges

### LINKED ISSUE #440: Use Babylon ESTree plugin

danez:
Fixes #440

### DIFF
--- a/babylon-to-espree/convertComments.js
+++ b/babylon-to-espree/convertComments.js
@@ -8,10 +8,5 @@ module.exports = function (comments) {
     } else if (comment.type === "CommentLine") {
       comment.type = "Line";
     }
-    // sometimes comments don't get ranges computed,
-    // even with options.ranges === true
-    if (!comment.range) {
-      comment.range = [comment.start, comment.end];
-    }
   }
 };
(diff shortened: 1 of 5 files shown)


hzoo:
we might be able to remove some other stuff in babylon-to-estree

danez:
Wasn't sure, a lot of other stuff is unrelated to estree, tokens, types. But yeah maybe there is some more.

hzoo:
maybe 

like https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/index.js#L32-L33
https://github.com/babel/babel-eslint/blob/master/babylon-to-espree/convertComments.js#L14

hzoo:
can land when it passes, can iterate

hzoo:
lol nvm? we can land what you have

danez:
Seems we don't add ranges to comments🤔

hzoo:
would be nice to have a perf check for this change (or this vs regular parser)

PR #523, opened 2017-09-28
Allow eslint to define the babylon plugins

kesne:
If you want to use `babel-eslint`, currently that means opting into every syntax feature that babylon offers. This allows you to provide your own array of babylon plugins in the eslint `parserOptions`. For us, this will allow us to enable stage-3 syntax features, but still make things like do-expressions a syntax error. If you don't define plugins, it'll default to enabling them all.

### DIFF
--- a/index.js
+++ b/index.js
@@ -392,7 +392,7 @@ exports.parseNoPatch = function(code, options) {
     allowSuperOutsideMethod: true,
     ranges: true,
     tokens: true,
-    plugins: [
+    plugins: options.plugins || [
       "flow",
       "jsx",
       "estree",


kesne:
Yeah, I think there's a path forward there. If we land https://github.com/babel/babel/pull/6332 and add a `babel-core` dependency, we should be able to run an empty transform when we start linting, then read the `parserOptions.plugins` and pass that to babylon. That way, `babel-eslint` will set itself up based on your `.babelrc` as expected.

kaicataldo:
I agree with @ljharb. Since syntax plugins aren't configurable by the user, I don't think we should force users to keep track of this configuration for themselves. I'm going to try to find some time to look into this.

kaicataldo:
WIP PR is open here: https://github.com/babel/babel-eslint/pull/711

All feedback welcome :)

kaicataldo:
https://github.com/babel/babel-eslint/pull/711 has now landed

hzoo:
Going to close since ^ landed!

PR #527, opened 2017-10-11
Allow newer versions of babel dependencies

danez:


hzoo:
Should we switch to scoped here?

hzoo:
I'l just do that

Issue #537, opened 2017-11-09
Template string with object with template string inside


scastiel:
The parsing fails when parsing a piece of code containing:
* a template string
* in this template string: an object
* in this object: another template string

Example: 
```js
`${ { a:`${2}` } }`
```

The inner template string is not converted correctly when passing into `convertTemplateType`.

I forked the repo and added a new test that currently fails to reproduce the problem:  https://github.com/scastiel/babel-eslint/blob/master/test/babel-eslint.js#L157-L159

PR #538, opened 2017-11-10
Fix parsing of nested templates with objects (issue #537)

scastiel:
To fix the issue (see #537) I had to refactor the `convertTemplateType` function. Basically the problem was that when an embedded expression contained braces, and in this braces another template with embedded expression (etc.), the current number of opened braces was not right.

Indeed storing the current number of opened braces is not sufficient, since in the outer expression we can have an opening brace, then a template string, then the matching closing brace. This means that we need to store the number of braces for a current context, and when we encounter a new embedded expression, create a new clean context. When going back to the initial expression, we take back the matching context.

To do that I chose to store a *context stack*: the top context is always the current one, and depending of the tokens we encounter, we push and pop new contexts.

Sorry if this isn't very clear; it's was very painful for me to understand where the problem came from, and it's still difficult to explain 😉

I would have been so proud to produce a code that was understandable by itself, but I'm forced to admit that the algorithm is quite complicated to understand, so I had to document the code a lot. I hope and think it's easier to understand and maintain than before.

### DIFF
--- a/babylon-to-espree/convertTemplateType.js
+++ b/babylon-to-espree/convertTemplateType.js
@@ -1,25 +1,53 @@
 "use strict";
 
 module.exports = function(tokens, tt) {
-  var startingToken = 0;
-  var currentToken = 0;
-  var numBraces = 0; // track use of {}
-  var numBackQuotes = 0; // track number of nested templates
+  const contextStack = [{ inTemplate: false, numBraces: 0 }];
+  let contextIndex = 0;
+  const getContext = () => contextStack[contextIndex];
 
-  function isBackQuote(token) {
-    return tokens[token].type === tt.backQuote;
-  }
+  for (let index = 0; index < tokens.length; index++) {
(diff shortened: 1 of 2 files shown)


hzoo:
espree has similar logic for templates since it uses acorn -> esprima output https://github.com/eslint/espree/blob/master/lib/token-translator.js#L43 not I guess my original implementation had a bug that espree didn't? I do wish we could just share it if that was possible @not-an-aardvark @mysticatea

mysticatea:
@hzoo Thank you. I checked with espree, then it worked correctly in this case. I have not understood espree's translation logic...

scastiel:
@hzoo @mysticatea The original issue I had was using *eslint-plugin-react* (see issue https://github.com/yannickcr/eslint-plugin-react/issues/1523). The problem occurred with *babel-eslint*, not with the default parser. I admit I don't understand how *espree* token translator works too 😕

PR #556, opened 2017-12-24
Make 2018 the default ecmaVersion for rules relying on parserOptions

kaicataldo:
Some core rules (and potentially custom rules in the ecosystem) rely on checking `parserOptions` in the rules themselves. Technically, since Babylon always parses the latest (including experimental) syntax, this should be set to the latest ESLint supports (currently 2018).

### DIFF
--- a/test/babel-eslint.js
+++ b/test/babel-eslint.js
@@ -37,7 +37,7 @@ function parseAndAssertSame(code) {
     range: true,
     comment: true,
     attachComment: true,
-    ecmaVersion: 8,
+    ecmaVersion: 2018,
     sourceType: "module",
   });
   var babylonAST = babelEslint.parseForESLint(code, {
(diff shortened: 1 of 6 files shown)


not-an-aardvark:
I don't think this has any effect -- ESLint [copies the parserOptions object](https://github.com/eslint/eslint/blob/cbf0fb9b209dd2d3efe69510925805da5fa8e29f/lib/linter.js#L543-L552) before passing it to parsers.

mysticatea:
Oh, I see. I had forgotten it.

But, I think that it's given to `eslint-scope`.

- https://github.com/babel/babel-eslint/pull/559/files#diff-3ab10361493aeb949966224850773473R320
- https://github.com/babel/babel-eslint/pull/559/files#diff-2d0b8df824f212d55b29824f29bf4b48R364

test/babel-eslint.js
```diff
@@ -71,7 +71,7 @@ function parseAndAssertSame(code) {
   // assert.equal(esAST, babylonAST);
 }
 
-describe("babylon-to-esprima", () => {
+describe("babylon-to-espree", () => {
```
kaicataldo: I think this is correct? I'm assuming this is a holdover from a long time ago :)

Issue #558, opened 2017-12-24
Incompatibility with eslint 4.14.0


ndelangen:
See: https://github.com/eslint/eslint/issues/9767

Something changed in esline `4.14.0` that is causing an incompatibility. With the result of:
```
> $ yarn lint                                                                          [±master ●]
yarn run v1.3.2
$ yarn lint:js . && yarn lint:md .
$ cross-env NODE_ENV=production eslint --cache --cache-location=.cache/eslint --ext .js,.jsx,.json .
Cannot read property 'type' of undefined
TypeError: Cannot read property 'type' of undefined
    at isForInRef (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:410:24)
    at variable.references.some.ref (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:447:21)
    at Array.some (<anonymous>)
    at isUsedVariable (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:446:40)
    at collectUnusedVariables (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:569:26)
    at collectUnusedVariables (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:576:17)
    at Program:exit (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/rules/no-unused-vars.js:621:36)
    at listeners.(anonymous function).forEach.listener (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/util/safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.leaveNode (/Users/dev/Projects/GitHub/storybook/core/node_modules/eslint/lib/util/node-event-generator.js:303:14)
[...]

PR #559, opened 2017-12-25
Prevent parseForESLint() behavior from changing after parse() is called

not-an-aardvark:
(fixes https://github.com/babel/babel-eslint/issues/558, fixes https://github.com/eslint/eslint/issues/9767)

This updates the `parseForESLint` implementation to still provide custom `visitorKeys`/`scopeManager` objects even after `parse()` has been called before (and `eslint-scope` has been monkeypatched).

~~The custom `scopeManager` should not be affected by the legacy monkeypatching to `eslintScope.analyze` and `estraverse.VisitorKeys`, because it does not use either of them.~~


not-an-aardvark:
Actually, this is probably not correct (see https://github.com/babel/babel-eslint/issues/558#issuecomment-353821396). Closing.

lib/analyze-scope.js
```diff
@@ -308,6 +308,10 @@ class Referencer extends OriginalReferencer {
 }
 
 module.exports = function(ast, parserOptions) {
+  if (OriginalReferencer._babelEslintPatched) {
+    return escope.analyze(ast, parserOptions);
+  }
+
   const options = {
```
mysticatea: I think `escope.analyze` cannot understand `parserOptions`. It needs this `options`.

not-an-aardvark: I agree, but just passing in that `options` object causes some additional errors. I think this needs to use the same option-translation logic that `parse-with-patch` uses.

not-an-aardvark: It seems like the monkeypatching in `parse-with-patch` relies on a global `eslintOptions` object, which isn't being set properly here because the `parse-with-patch` function isn't being called. To fix it, I think it would be better to pass the options down directly.

mysticatea: I'm investigating it as well. But I have to leave while an hour.

not-an-aardvark:
I think I fixed the issue. I also updated the regression tests to check both old and new versions of ESLint.

I realized that it was necessary for `babel-eslint` to use a global when monkeypatching `eslint-scope`, because it can't modify `parserOptions` (since ESLint clones them [here](https://github.com/eslint/eslint/blob/cbf0fb9b209dd2d3efe69510925805da5fa8e29f/lib/linter.js#L543-L552)), and ESLint would always default to `sourceType: script` and `ecmaVersion: 5` [here](https://github.com/eslint/eslint/blob/cbf0fb9b209dd2d3efe69510925805da5fa8e29f/lib/linter.js#L514-L526).

Unfortunately, this makes `eslint-scope` effectively useless for all other modules after `babelEslint.parse` is called, because it doesn't use the provided options. (Actually, this was already a problem in previous versions of ESLint if another parser was being used at the same time as `babel-eslint`.) Separately from this change, maybe it would be a good idea to release a major version of `babel-eslint` that removes all of the monkeypatching, so that it doesn't break other parsers when `eslint-plugin-import` causes it to monkeypatch things.

package.json
```diff
@@ -39,9 +39,11 @@
   "devDependencies": {
     "babel-eslint": "^8.0.0",
     "dedent": "^0.7.0",
-    "eslint": "^4.12.1",
+    "eslint": "^4.14.0",
     "eslint-config-babel": "^7.0.1",
+    "eslint-old": "npm:eslint@4.13.1",
```
mysticatea: Wow, I learned this notation today.

not-an-aardvark: Unfortunately, I think it's only supported by `yarn` and not `npm`.

mysticatea: Oh, thank you..

mysticatea:
I have a concern about tests. Most test cases look to be run after monkey-patched. I'm writing improvement...

mysticatea:
I updated this PR about two things:

- I separated monkey-patching to `lib/patch-eslint-scope.js` then made it applies the patch of `escope.analyze` for each `parserOptions`. Then `eslintOptions` global state was removed.
- I modified `npm test` script to test three cases properly:
  - `eslint@4.14.0` normally. (`mocha`)
  - `eslint@4.14.0` after patched. (`mocha --require test/fixtures/preprocess-to-patch.js`)
  - `eslint@4.13.1`. (`mocha --require test/fixtures/use-eslint-old.js`)

@not-an-aardvark Please check it.

test/babel-eslint.js
```diff
@@ -539,31 +498,4 @@ describe("Public API", () => {
       babelEslint.parseNoPatch("foo", {})
     );
   });
-
-  /*
-   * This test ensures that the enhanced referencer does not get used if eslint-scope has already been
-   * monkeypatched, because this causes some correctness issues. For example, if the enhanced referencer
-   * is used after the original referencer is monkeypatched, type annotation references are counted twice.
-   */
-  it("does not visit type annotations multiple times after monkeypatching and calling parseForESLint()", () => {
```
mysticatea: I moved this test to `test/z_parser-for-eslint-after-patched.js` in order to not call `babelEslint.parse` before other tests.

lib/patch-eslint-scope.js
```diff
@@ -0,0 +1,376 @@
...
+    try {
+      const modules = getModules();
+      monkeypatch(modules);
+
+      // Store to patch for each call.
+      escope = modules.escope;
+      escopeAnalyze = modules.escope.analyze;
+    } catch (err) {
+      console.error(err.stack);
+      process.exit(1);
```
not-an-aardvark: I don't think calling `process.exit(1)` here is a good idea, because another module could be doing something on the process. Throwing an error seems like it would be better instead.

mysticatea: It's just copied from https://github.com/babel/babel-eslint/pull/559/commits/db14da01f2751053e1f911ad344c2966ea58e048#diff-3fa2324e5c70e869dd8b91ba47d140d1L379
But, I agree `process.exit(1)` is bad.

not-an-aardvark: Oh, I hadn't realized that. In that case I suppose we don't have to change it now.

mysticatea: removed.

not-an-aardvark:
LGTM

mysticatea:
Thank you for all your work!

PR #562, opened 2017-12-25
Fix: add Literal type to visitorKeys

mysticatea:
Refs https://github.com/babel/babel-eslint/issues/558#issuecomment-353861903.

Babel's AST did not have `Literal` type.
This PR adds it to improve performance.

~~Though I'm not sure why `Property` is unknown in that debug log... I confirmed it's `"Property": [ 'decorators', 'key', 'value' ]`.~~

### DIFF
--- a/lib/parse-with-scope.js
+++ b/lib/parse-with-scope.js
@@ -1,13 +1,12 @@
 "use strict";
 
-const t = require("@babel/types");
+const visitorKeys = require("./visitor-keys");
 const analyzeScope = require("./analyze-scope");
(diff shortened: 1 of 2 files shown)


mysticatea:
Also, `visitorKeys` in `./parse-with-scope.js` was wrong, so some node types (Literal, Property, MethodDefinition) have used the fallback visitor keys.

hzoo:
👍, Fyi we have https://github.com/babel/babel/blob/master/packages/babylon/README.md#output

PR #565, opened 2018-01-02
update babel packages

jony89:
There is some critical bug fixes in version babylon#7.0.0-beta.33 regard `async` - `await` syntax which is needed pretty immediately :)
 
also removing one test as `import type *` is invalid, since the namespace can't be a type, and it is shown as an error now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,10 +11,10 @@
     "url": "https://github.com/babel/babel-eslint.git"
   },
   "dependencies": {
-    "@babel/code-frame": "7.0.0-beta.31",
-    "@babel/traverse": "7.0.0-beta.31",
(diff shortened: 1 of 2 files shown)


jony89:
@mysticatea any idea why this is failing ?

existentialism:
@jony89 thanks for the PR. In the latest Babel version, we now match flow's behavior with making the following a syntax error:

`import type * as namespace from 'bar';`

The fix is to remove the following test:
https://github.com/babel/babel-eslint/blob/bf9092ab6c3bccf5c9d4cbc91de59d7edfe3641e/test/non-regression.js#L1082

jony89:
@mysticatea could u merge this please ? ty :)

PR #577, opened 2018-01-15
Fix: wrong token type of ! and ~ (fixes #576)

### LINKED ISSUE #576: incorrect token type for negated regexp
See https://github.com/vuejs/eslint-plugin-vue/issues/349

**Original code:**
```js
return (
  // it's inside node_modules
  /node_modules/.test(module.context) &&
  // and not a CSS file (due to extract-text-webpack-plugin limitation)
  !/\.css$/.test(module.request)
)
```

**The token passed into `isComment`:** (it's the `!` before the regexp)
```json
{
  "type": {
    "label": "!",
    "beforeExpr": true,
    "startsExpr": true,
    "rightAssociative": false,
    "isLoop": false,
    "isAssign": false,
    "prefix": true,
    "postfix": false,
    "binop": null,
    "updateContext": null
  },
  "value": "!",
  "start": 963,
  "end": 964,
  "loc": {
    "start": {
      "line": 26,
      "column": 10
    },
    "end": {
      "line": 26,
      "column": 11
    }
  },
  "range": [
    963,
    964
  ]
}
```

> As ESLint AST spec, the type must be a string
&mdash; mysticatea

mysticatea:
Fixes #576.

This PR fixes a bug of token types.

~~The token type of bitwise/logical NOT should be `tt.prefix`, but the `tt.prefix` does not match to bitwise/logical NOT in the current Babylon parser. This PR adds a workaround.~~

EDIT: Babylon does not have `tt.prefix`. It has `tt.bang` and `tt.tilde` instead.

### DIFF
--- a/lib/babylon-to-espree/toToken.js
+++ b/lib/babylon-to-espree/toToken.js
@@ -38,7 +38,8 @@ module.exports = function(token, tt, source) {
     type === tt.plusMin ||
     type === tt.modulo ||
     type === tt.exponent ||
-    type === tt.prefix ||
+    type === tt.bang ||
+    type === tt.tilde ||
     type === tt.doubleColon ||
     type.isAssign
   ) {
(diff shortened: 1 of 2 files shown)


mysticatea:
Oops, I'm wrong. `tt.prefix` does not exist in babylon.

KaelWD:
When is this likely to be released?

PR #584, opened 2018-01-21
Support cyclic type declarations (#485)

### LINKED ISSUE #485: Circular flow type definitions trigger no-use-before-define
Brought here from https://github.com/eslint/eslint/issues/8703, by @soda0289 request:
When using core rule `no-use-before-define`, default for example in `standard`, it is applied to type definitions.
This could be a good thing, but the current implementation gives an unavoidable error on circular definitions:

```js
/* @flow */
type File = {chunks: Array<Chunk>}
type Chunk = {file: File}
```
**produces**
```
.../src/commands.js
  2:21  error  'File' was used before it was defined  no-use-before-define

✖ 1 problem (1 error, 0 warnings)
```
**OR**
```js
/* @flow */
type Chunk = {file: File}
type File = {chunks: Array<Chunk>}
```
**produces**
```
.../src/commands.js
  2:28  error  'Chunk' was used before it was defined  no-use-before-define

✖ 1 problem (1 error, 0 warnings)
```
respectively.

Currently, disabling `no-use-before-define` in the files that have flow enabled (all for me) is a fix when actual flow is checked, since that catches undefined.
However, it is still probably a bug, since:
```js
// This is an actual error
const a = {b}
const b = {a}
// While this works and is often necessary
type A = {b:B}
type B = {a:A}
```

joa:
Using a type before its definition shouldn't trigger a no-use-before-define warning. Flow types should experience them same behavior as function declarations (they behave currently more like function expressions).

Cyclic type dependencies are common when using flow.
For instance: 
```
type Node<T> = { head: T; tail: Node<T> }
```

Fixes #485

### DIFF
--- a/lib/babylon-to-espree/toAST.js
+++ b/lib/babylon-to-espree/toAST.js
@@ -72,6 +72,15 @@ var astTransformVisitor = {
       delete node.bound;
     }
 
+    if (path.isTypeAlias()) {
+      node.type = "FunctionDeclaration";
+      node.generator = false;
+      node.async = false;
+      node.expression = false;
+      node.params = [];
+      node.body = node.right;
+    }
+
     // flow: prevent "no-undef"
     // for "Component" in: "let x: React.Component"
     if (path.isQualifiedTypeIdentifier()) {
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -1157,6 +1157,18 @@ describe("verify", () => {
         []
       );
     });
+
+    it("cyclic type dependencies #485", () => {
+      verifyAndAssertMessages(
+        unpad(`
+          type Node<T> = { head: T, tail: Node<T> };
+          type A = B[];
+          type B = number;
+        `),
+        { "no-use-before-define": 1 },
+        []
+      );
+    });
   });
 
   it("class usage", () => {


xtuc:
Sorry for the delay, we need to review your changes before we can merge this.

lib/babylon-to-espree/toAST.js
```diff
@@ -72,6 +72,15 @@ var astTransformVisitor = {
...
     }
 
+    if (path.isTypeAlias()) {
+      node.type = "FunctionDeclaration";
+      node.generator = false;
+      node.async = false;
+      node.expression = false;
+      node.params = [];
+      node.body = node.right;
+    }
```
xtuc: Do we have some tests for the babylon-to-espress transform? That is surprising to me that this don't have any side effect?

joa: https://github.com/babel/babel-eslint/blob/master/test/babel-eslint.js#L68

Shuliyey:
:+1:

mdsib:
Thanks for this! Is there anything stopping the merge?

mourner:
@xtuc hey Sven, any news about this PR?

hzoo:
Sorry, haven't been able to look at all this in a while! Thanks for the PR.

hzoo:
Ok weird the test is failing on master

pnevyk:
This change (generally) breaks eslint plugins that use `TypeAlias` visitor. To me, changing node type from `TypeAlias` to `FunctionDeclaration` doesn't look like a correct solution (however, I have no insight into issue that is fixed by this). Is this going to be reverted?

hzoo:
Yeah I think we will need to revert, was trying to look into the solution earlier

Issue #573, opened 2018-01-09
Feature: Enable syntax based on reading from `.babelrc`


hzoo:
Ref https://github.com/babel/babel-eslint/issues/434 (subset of syntax), https://github.com/babel/babel-eslint/pull/523 (user plugins), https://github.com/babel/babel-eslint/issues/505 (Typescript)

https://github.com/babel/babel-eslint/blob/bf27f6021e3a3661589bfa1e314f45876edeb20b/lib/parse.js#L18-L22

Instead of having to:
- add to the plugin list whenever there is a new syntax and release a new version
- using a wildcard/star `*` option (found out that was a bad idea due to breaking changes between plugins like decorators/decorators2 so wouldn't be able to easily resolve that)
- [expose a user config whitelist](https://github.com/babel/babel-eslint/pull/523)

We should be able to just read from a user's Babel config. `@babel/core` should expose a way for another tool to get relevant information from a user's config file (like return the list of syntax plugins used). `babel-eslint` could try to read that and just populate the parser options and we can remove the defaults. I think it's safe to assume if you are using babel-eslint you are using babel, and should have a config (can wrap in try/catch or fallback to default list)?

- Less config (re-using config)
- Enable subset of features
- We should be able to remove all deps and just do a peerDep on `@babel/core` maybe? Then it would use the version of babel in your project rather than it's own dep. And maybe the same for the eslint deps?

https://github.com/babel/babel-eslint/blob/bf27f6021e3a3661589bfa1e314f45876edeb20b/package.json#L14-L17

edit: I believe we are only using `@babel/types` to get the the VISITOR_KEYS, and traverse to do a fast transform of the AST

Thoughts? cc @ljharb @kaicataldo @loganfsmyth @azz @danez @zertosh

azz:
Makes perfect sense to me. Much better long term solution and it's good for perf, too.

kaicataldo:
I'd like to work on this.

kaicataldo:
And agree, I think it should use the new `@babel/core` API and fallback to the default (potentially outdated) list when it isn't there.

hzoo:
Awesome, looks like @loganfsmyth was thinking about the babel side in https://github.com/babel/babel/pull/6332#issuecomment-333399009 (didn't realize we had a PR for getting parserOpts earlier)

ljharb:
[...]

PR #594, opened 2018-02-20
Use @babel/core#parse

kaicataldo:
fixes https://github.com/babel/babel-eslint/issues/573

This is a WIP. I still need to figure out how to test this.

@babel/babel Would love input on if this seems like a good way to go, behavior-wise.


package.json
```diff
@@ -36,7 +32,14 @@
     "url": "https://github.com/babel/babel-eslint/issues"
   },
   "homepage": "https://github.com/babel/babel-eslint",
+  "peerDependencies": {
```
ljharb: This is a lot of peer deps; eslint and babel-core makes sense, but aren’t the other two addressed by babel-core?

package.json
```diff
@@ -36,7 +32,14 @@
...
+    "@babel/code-frame": ">=7.0.0-beta.40",
+    "@babel/core": ">=7.0.0-beta.40",
+    "babylon": ">=7.0.0-beta.40",
+    "eslint": ">=4.14.0"
+  },
   "devDependencies": {
+    "@babel/core": "7.0.0-beta.40",
```
ljharb: The devDep range should be identical to the peerDep range.

lib/parse.js
```diff
@@ -1,87 +1,73 @@
...
+    parserOpts: {
+      sourceType: options.sourceType,
+      allowImportExportEverywhere: options.allowImportExportEverywhere, // consistent with Espree
+      allowReturnOutsideFunction: true,
+      allowSuperOutsideMethod: true,
+      ranges: true,
+      tokens: true,
+    },
   };
+  var enableCodeFrame = options.hasOwnProperty("codeFrame") ? options.codeFrame : true;
```
ljharb: This might break if i pass a “hasOwnProperty” option; you probably want `Object.prototype.hasOwnProperty.call` or the `has` module (i realize this isn’t related to your PR, but it’s an easy fix)

package.json
hzoo: Yeah code-frame and babylon are dependencies of babel/core that's true. So if you are on npm 3 it should be fine?

kaicataldo: Since we rely on those packages directly and they could potentially not be required by @babel/core I thought it would be safer this way. Can definitely remove them if we think that's being overprotective!

ljharb: The goal is to set up the deps such that when the user runs `npm install && npm ls`, everything exits zero.

Since we rely on them directly, I'd rather them be deps. *However* we could make them *both* deps and peerDeps, and then anyone on npm3+ should get them autoinstalled and satisfy the peer dependency, and any conflicts due to changes in babel-core would cause `npm ls` to fail.

lib/parse.js
```diff
@@ -1,87 +1,73 @@
...
-      "optionalChaining",
-      "importMeta",
-      "classPrivateProperties",
-      "bigInt",
-      "optionalCatchBinding",
-      "throwExpressions",
-      "pipelineOperator",
-      "nullishCoalescingOperator",
-    ],
+    filename: (options.filePath && options.filePath !== ESLINT_DEFAULT_FILEPATH) ? options.filePath : '',
```
loganfsmyth: I'd say pass `undefined` if there is no filename.

lib/parse.js
```diff
@@ -1,87 +1,73 @@
...
-                column: err.column,
+          // remove trailing "(LINE:COLUMN)" acorn message and add in esprima syntax error message start
+          err.message =
+            "Line " +
+            err.lineNumber +
+            ": " +
+            err.message.replace(/ \((\d+):(\d+)\)$/, "") +
+            // add codeframe
+            "\n\n" +
+            codeFrameColumns(
```
loganfsmyth: Doesn't `.parse` throw with an included code frame now?

lib/parse.js
```diff
@@ -1,87 +1,73 @@
 "use strict";
 
 var babylonToEspree = require("./babylon-to-espree");
-var parse = require("babylon").parse;
 var tt = require("babylon").tokTypes;
```
loganfsmyth: Should we consider exporting the `babylon` module namespace from `babel-core`? The only reason not to is if we were going to version them independently, and realistically that doesn't seem likely.

loganfsmyth: Then we could ditch the `babylon` dependency and not risk them being out of sync, right?

phyllisstein:
👋Is there anything the user community can do to support this work? I'd love to see TypeScript support land, so I'd be happy to volunteer some nights and/or weekends if I can be useful.

kaicataldo:
@phyllisstein TypeScript support is a separate issue than this PR solves. Hoping I can pick this back up soon - I've been busy with work and just haven't had time.

seanCodes:
Any developments on this? What’s left to be done? I’m also happy to help if I can.

kaicataldo:
Sorry for the hold up. Hoping I can find time for this in the next few weeks.

dan-kez:
@kaicataldo - with the recent approval from @sibelius is there a chance we can resolve the conflicts for merge? It would be great to use this feature.

kaicataldo:
Closing this in favor of starting a new PR. Will start looking at it tonight.

PR #607, opened 2018-04-09
updates readme for latest eslint & babel-eslint

aoberoi:
while i'm not 100% confident that the version range (`>=6.x`) is correct for compatibility with the latest versions of ESLint, it was def missing and unclear whether ESLint v4.x was supported. anecdotally, these versions work for me. also the installation instructions were referencing older versions of these packages (IMHO they don't need to reference any version at all, but i chose to stay with the convention).


hzoo:
We should probably remove the old versions anyway, thanks!

PR #640, opened 2018-06-23
Revert bump to babel 51

existentialism:


hzoo:
Forgot that in beta we dropped Node 4, so having a non-beta version drop it automatically was unintentional. Yeah we'll revert here with a patch and do a new version.

I guess `9.0.0-beta.0`? We could also switch to using a peerDep on `@babel/core` instead

TrySound:
We can't consume new syntax features. Would be good if you will release beta.

existentialism:
@TrySound published v9.0.0-beta.1!

TrySound:
Awesome! Thank you

PR #642, opened 2018-06-27
Upgraded Babel to 7.0.0-beta.51, with changes to decorators

rubennorte:
I upgraded the Babel dependencies to the latest beta (with support for new Flow syntax) and fixed all failing tests.

The most relevant change is that Babel now supports the new decorators proposal by default, so we need to switch between the new and the old proposal. **This is a breaking change**.

To enable the legacy decorators proposal users should add a specific parser option:

```javascript
{
  parserOptions: {
    ecmaFeatures: {
      legacyDecorators: true
    }
  }
}
```


existentialism:
@rubennorte published as v9.0.0-beta.1, thanks!

PR #645, opened 2018-07-05
Support new flow syntax in scope analysis

### LINKED ISSUE #644: Flow types used in explicit type arguments in function calls are reported as not being used
Given the following code:

```
type Type = string;

function identity<T>(val: T): T {
  return val;
}

const foo = identity<Type>('foo');
```

`Type` is being reported as not being used by ESLint. Identifiers used inside explicit type arguments in function calls should be visited and not reported as unused.

rubennorte:
~Fixes #644.~
~Requires https://github.com/babel/babel/pull/8273 to be merged and released before merging this.~

This is just a test case that failed before the upgrade.

### DIFF
--- a/test/non-regression.js
+++ b/test/non-regression.js
@@ -564,6 +564,17 @@ describe("verify", () => {
       );
     });
 
+    it("polymorphic/generic types - function calls #644", () => {
+      verifyAndAssertMessages(
+        `
+          import type {Type} from 'Type';
+          function f<T>(): T {}
+          f<Type>();
+        `,
+        { "no-unused-vars": 1, "no-undef": 1 }
+      );
+    });
+
     it("support declarations #132", () => {
       verifyAndAssertMessages(
         `


lib/analyze-scope.js
```diff
@@ -18,6 +18,10 @@ const flowFlippedAliasKeys = t.FLIPPED_ALIAS_KEYS.Flow.concat([
   "ObjectPattern",
   "RestElement",
 ]);
+
+// TODO Remove!
+t.VISITOR_KEYS.CallExpression.push("typeArguments");
```
rubennorte: I'll remove this once my other PR in Babel gets merged and released. Please do not merge this yet as it is now.

lib/analyze-scope.js
```diff
@@ -131,6 +131,11 @@ class Referencer extends OriginalReferencer {
     super.visitProperty(node);
   }
 
+  CallExpression(node) {
```
rubennorte: This is not needed anymore after the Babel upgrade. I'll just leave the test case.

PR #689, opened 2018-09-25
Drop old monkeypatching behavior

existentialism:
An alternative to adding a direct dep on estraverse (https://github.com/babel/babel-eslint/pull/685), let's just drop the old monkeypatching behavior.

/cc: @mysticatea @kaicataldo


hzoo:
Ref https://github.com/babel/babel-eslint/pull/685
Old PR: https://github.com/babel/babel-eslint/pull/542

PR #696, opened 2018-09-27
Add support for opaque types in scope analysis

pnevyk:
Opaque type is virtually the same as type alias from syntax point of view, so I just copied the code for TypeAlias visitor and changed `right` to `impltype`.


kaicataldo:
Is this still something we want? If so, can we add some tests?

pnevyk:
Closing in favor of #750

PR #706, opened 2018-10-27
 Update ESLint config

kaicataldo:
Looking to pick up https://github.com/babel/babel-eslint/pull/594 again and saw a few things that I thought should be updated in a separate PR.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,12 +5,22 @@ module.exports = {
     "prettier"
   ],
   rules: {
-    "no-var": 0,
-    "max-len": 0,
+    "max-len": "off",
+    "strict": "error",
     "prettier/prettier": "error",
   },
   env: {
     node: true,
-    mocha: true
-  }
+  },
+  parserOptions: {
+    sourceType: "script",
+  },
+  overrides: [
+    {
+      files: ["test/**/*"],
+      env: {
+        mocha: true
+      }
+    }
+  ]
 };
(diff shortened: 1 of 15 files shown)


.eslintrc.js
```diff
@@ -5,12 +5,22 @@ module.exports = {
     "prettier"
   ],
   rules: {
-    "no-var": 0,
-    "max-len": 0,
+    "max-len": "off",
+    "strict": "error",
```
kaicataldo: I noticed some files were using directives and some weren't, and I'm assuming we want to be in strict mode for all files.

.eslintrc.js
```diff
@@ -5,12 +5,22 @@ module.exports = {
...
     "prettier/prettier": "error",
   },
   env: {
     node: true,
-    mocha: true
-  }
+  },
+  parserOptions: {
+    sourceType: "script",
```
kaicataldo: Needed to override https://github.com/babel/eslint-config-babel/blob/master/index.js#L7 for the `strict` rule

.eslintrc.js
```diff
@@ -5,12 +5,22 @@ module.exports = {
...
+  },
+  overrides: [
+    {
+      files: ["test/**/*"],
+      env: {
+        mocha: true
```
kaicataldo: Scoping this env just to tests so that we don't accidentally use these globals in the source code.

PR #709, opened 2018-10-30
fix index path typo

emersonlaurentino:
by default github uses `src` as the base path, but `lib` is used for this case.


kaicataldo:
LGTM - thanks for contributing!

PR #563, opened 2017-12-28
Clarify what all valid code means in readme

the-spyke:
Fixes #555

PR #711, opened 2018-11-07
Use @babel/core#parse

kaicataldo:
refs https://github.com/babel/babel-eslint/issues/573

This PR updates babel-eslint to require `@babel/core` as a peer dependency and to use the user's Babel configuration when parsing files for ESLint. Up until now, babel-eslint has manually enabled all plugins (with the list falling out of date frequently). It also meant that babel-eslint could parse syntax that a configured instance of Babel itself didn't allow at compile time.

This change makes it so that the same Babel configuration is used for the parser both for linting and for compilation, and should eliminate a whole class of bugs as well as give the end user more fine-grained control over what the parser can and cannot parse.

Note that this is a breaking change, as using babel-eslint will now require `@babel/core@>=7.2.0` and a valid Babel configuration file to run (see https://github.com/babel/babel-eslint/pull/711#issuecomment-446647087 for some discussion around this).

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -13,7 +13,7 @@ module.exports = function(code, options) {
     envName: options.babelOptions.envName,
     configFile: options.babelOptions.configFile,
     babelrc: options.babelOptions.babelrc,
-    babelrcRoots: options.babelOptions.babelrRoots,
+    babelrcRoots: options.babelOptions.babelrcRoots,
     extends: options.babelOptions.extends,
     env: options.babelOptions.env,
     overrides: options.babelOptions.overrides,


lib/parse.js
```diff
@@ -1,54 +1,47 @@
 "use strict";
 
 const babylonToEspree = require("./babylon-to-espree");
-const parse = require("@babel/parser").parse;
+const { parseSync: parse, traverse } = require("@babel/core");
 const tt = require("@babel/parser").tokTypes;
```
kaicataldo: @loganfsmyth You had mentioned [exporting `@babel/parser` in `@babel/core`](https://github.com/babel/babel-eslint/pull/594#discussion_r169840591). This makes a lot of sense to me, and it would allow us to only have `@babel/core` as a dependency.

ljharb: That would be awesome, and would remove the need to import it [here](https://github.com/airbnb/babel-plugin-inline-react-svg/blob/master/src/index.js#L3), which would have allowed simultaneous babel 6 and 7 compat, and may allow for simultaneous 7 and 8+ compact moving forwards.

README.md
```diff
@@ -81,15 +81,19 @@ Check out the [ESLint docs](http://eslint.org/docs/rules/) for all possible rule
 
 - `sourceType` can be set to `'module'`(default) or `'script'` if your code isn't using ECMAScript modules.
 - `allowImportExportEverywhere` (default `false`) can be set to `true` to allow import and export declarations to appear anywhere a statement is allowed if your build environment supports that. Otherwise import and export declarations can only appear at a program's top level.
+- `ecmaFeatures.globalReturn` (default `false`) allow return statements in the global scope when used with `sourceType: "script"`.
 
-**.eslintrc**
+**.eslintrc.json**
```
ljharb: why tho; .eslintrc remains the defacto standard in the community :-p

README.md
```diff
@@ -81,15 +81,19 @@ Check out the [ESLint docs](http://eslint.org/docs/rules/) for all possible rule
...
 
 ```json
 {
   "parser": "babel-eslint",
   "parserOptions": {
     "sourceType": "module",
     "allowImportExportEverywhere": false,
+    "ecmaFeatures": {
+      "globalReturn": true
```
ljharb: why is this set to `true` and not the default, `false`?

kaicataldo: Ah, good catch. Will fix!

README.md
ljharb: tecnically yes, but in practice it'd be way too problematic to ever remove.

kaicataldo: Yep - that's why it's still supported :)

lib/parse.js
loganfsmyth: @kaicataldo The issue with doing this as-is, and the reason I think the other PR is important, is that there isn't any guarantee that the token typess returned from `parseSync` will actually be the same as `tokTypes` here. If the packages aren't deduped, they could be totally independent objects that are just deep-equal. The only way to be confident that the `tokTypes` values will match is to re-export them.

kaicataldo: Thanks for the explanation! That makes sense. Will update my other PR.

kaicataldo:
For anyone coming across this - I'm just waiting for https://github.com/babel/babel/pull/8986 to be merged and released to finish this up.

sibelius:
@kaicataldo it is merged 💯

tomwidmer:
@sibelius @kaicataldo And it is released https://github.com/babel/babel/releases/tag/v7.2.0 👍

kaicataldo:
Planning to pick this up again this week. Here's an interesting edge case that I don't think was considered with this change: https://github.com/eslint/eslint/issues/11189#issuecomment-446646882. Thoughts?

kentcdodds:
I can't tell from the PR. Is there a way to configure the location of the babel configuration as one of the `parserOptions` or something? With a toolkit (where the configuration does not necessarily live at the root of the project) being able to specify configuration directly would be really helpful.

kaicataldo:
@kentcdodds This was discussed elsewhere (trying to find it and am failing - might've been on Slack 😅) and I hadn't gotten around to it yet. Added [here](https://github.com/babel/babel-eslint/pull/711/files#diff-9f97a9c36fbbfb0d19068d6fae3bb84eR9). This was also useful for testing!

README.md
```diff
@@ -45,12 +49,12 @@ It just needs to export a `parse` method that takes in a string of code and outp
 
 ### Supported ESLint versions
 
-ESLint | babel-eslint
------------- | -------------
-4.x | >= 6.x
-3.x | >= 6.x
-2.x | >= 6.x
-1.x | >= 5.x
+| ESLint | babel-eslint |
```
kaicataldo: Added by Prettier

kaicataldo:
This is ready for review/discussion. I'd love to figure out how to make these tests easier to manage, but I think the implementation here is reasonable and we can iterate in the feature. Let me know how you feel.

Would love to get the team's feelings about throwing when the required version of `@babel/core` isn't found. Also, am I missing anything configuration-wise? I know a lot changed in babel@7 and I'm still not super familiar with all the options surrounding configs, particularly in monorepos.

villesau:
@kaicataldo any chance to get this merged, or updating babel versions in separate PR in the meantime? More recent babel have some syntax features that's needed when updating Flow.

lib/parse.js
```diff
@@ -1,54 +1,21 @@
...
-      "classPrivateProperties",
-      "bigInt",
-      "optionalCatchBinding",
-      "throwExpressions",
-      ["pipelineOperator", { proposal: "minimal" }],
-      "nullishCoalescingOperator",
-      "logicalAssignment",
-    ],
+    configFile: options.configFile,
+    filename: options.filePath,
```
loganfsmyth: I think we should probably pass through all the options that can directly influence a config:

* `cwd`
* `root`
* `rootMode`
* `envName`
* `configFile`
* `babelrc`
* `babelrcRoots`
* `extends`
* `env`
* `overrides`
* `test`
* `include`
* `exclude`
* `ignore`
* `only`

It would be great to also set
```
caller: {
  name: "babel-eslint",
},
```

kaicataldo: I have updated this, but I'm wondering if we do really want the config merging options here - shouldn't we be able to just provide the config loading options that would point to the config file that would contain the config merging options? This would avoid duplication in the `.eslintrc.js` and `babel.config.js` files.

lib/index.js
```diff
@@ -15,5 +36,7 @@ exports.parseForESLint = function(code, options) {
 };
 
 exports.parseNoPatch = function(code, options) {
+  options = options || {};
```
kaicataldo: This looks like an old API that is no longer used. Since we're making a breaking change here anyway, I made another PR to remove this method: https://github.com/babel/babel-eslint/pull/733

The changes here are just a stopgap to allow tests to pass in the interim.

lib/parse.js
kaicataldo: I personally would prefer to start with fewer rather than more options (especially since this project is already difficult to maintain), but am happy to defer to what you think is best. :+1:

README.md
```diff
@@ -3,54 +3,58 @@
...
 
-For questions and support please visit the [`#discussion`](https://babeljs.slack.com/messages/discussion/) babel slack channel (sign up [here](https://github.com/babel/notes/issues/38)) or eslint [gitter](https://gitter.im/eslint/eslint)!
+For questions and support please visit the [`#discussion`](https://babeljs.slack.com/messages/discussion/) Babel Slack channel (sign up [here](https://github.com/babel/notes/issues/38)) or the ESLint [Gitter](https://gitter.im/eslint/eslint).
 
-> Note that the `ecmaFeatures` config property may still be required for ESLint to work properly with features not in ECMAScript 5 by default. Examples are `globalReturn` and `modules`).
+## Why Use babel-eslint?
+
+You only need to use babel-eslint if you are using Babel to use types (Flow) or experimental ECMAScript features not supported in ESLint itself yet. If you are not using Babel to transform your code, please use the relevant parser for your chosen flavor of ECMAScript (note that the default parser supports all non-experimental syntax as well as JSX).
```
SimenB: Mention typescript?

README.md
```diff
@@ -3,54 +3,58 @@
...
+## Questions and support
 
-> If there is an issue, first check if it can be reproduced with the regular parser or with the latest versions of `eslint` and `babel-eslint`!
+If you have an issue, please first check if it can be reproduced with the default parser and with the latest versions of `eslint` and `babel-eslint`. If it is not reproducible with the default parser, it is most likely an issue with babel-eslint.
 
-For questions and support please visit the [`#discussion`](https://babeljs.slack.com/messages/discussion/) babel slack channel (sign up [here](https://github.com/babel/notes/issues/38)) or eslint [gitter](https://gitter.im/eslint/eslint)!
+For questions and support please visit the [`#discussion`](https://babeljs.slack.com/messages/discussion/) Babel Slack channel (sign up [here](https://github.com/babel/notes/issues/38)) or the ESLint [Gitter](https://gitter.im/eslint/eslint).
 
-> Note that the `ecmaFeatures` config property may still be required for ESLint to work properly with features not in ECMAScript 5 by default. Examples are `globalReturn` and `modules`).
+## Why Use babel-eslint?
```
SimenB: Weird with uppercase `U` in `Use`

README.md
kaicataldo: Thanks for the review! I was in the process of writing this stuff up - should be resolved now!

README.md
kaicataldo: Thanks for the review! I was in the process of writing this stuff up - should be resolved now!

jquense:
really excited about this 👍 looking forward to consolidating tslint into eslint :P

README.md
```diff
@@ -3,101 +3,105 @@
...
 
 ```sh
-$ npm install eslint@4.x babel-eslint@8 --save-dev
+$ npm install eslint babel-eslint --save-dev
 # or
-$ yarn add eslint@4.x babel-eslint@8 -D
+$ yarn add eslint babel-eslint -D
 ```
 
+**Note:** babel-eslint requires `babel/core@>=7.2.0` and a valid Babel configuration file to run. If you do have this already set up, please see the [Babel Usage Guide](https://babeljs.io/docs/en/usage).
```
kenneth-gray: Should this say `don't have this already setup` rather than `do have this already setup`?

lib/parse.js
```diff
@@ -1,54 +1,52 @@
...
+/*
+ * A filePath may not exist when the --stdin option is used on the command line or
+ * CLIEngine#executeOnText() is used programmatically. When it doesn't exist, ESLint sets
+ * the filePath to "<text>".
+ *
+ * https://eslint.org/docs/user-guide/command-line-interface#using-stdin
+ * https://eslint.org/docs/developer-guide/nodejs-api#cliengineexecuteontext
+ * https://github.com/eslint/eslint/blob/master/lib/cli-engine.js#L177
+ */
+const ESLINT_DEFAULT_FILEPATH = "<text>";
```
kaicataldo: I just came across [this](https://github.com/eslint/eslint/blob/master/lib/linter.js#L308) - going to talk to the team and make sure I'm understanding all the possible values we use when there isn't a filename provided. Seems confusing, so just want to make sure I understand the rationale here. Will update when I know more.

README.md
existentialism: :+1:

kaicataldo: Nice catch - thanks!

lib/parse.js
kaicataldo: Given that any path could be provided by a consumer, I think it's best to not try to guess the user's intent here and to just pass through what babel-eslint is given.

README.md
```diff
@@ -3,101 +3,107 @@
...
+module.exports = {
+  parser: "babel-eslint",
+  parserOptions: {
+    sourceType: "module",
+    allowImportExportEverywhere: false,
+    ecmaFeatures: {
+      globalReturn: false,
+    },
+    babelOptions: {
+      configFile: "path/to/config.js",
```
kentcdodds: This is fantastic. Very important for toolkits that encapsulate eslint config. Thank you for adding this feature 💯

cKiril:
Hey all, when can we expect this to roll out to production?

kaicataldo:
I've been continuing to do some manual testing, and have two things that I wanted to get feedback on (but I don't think they need to be blockers for landing this):

1. With the default ESLint results formatter, the `Parsing error: babel-eslint@10.0.1 does not support @babel/core@7.1.6. Please downgrade to babel-eslint@^10 or upgrade to @babel/core@>=7.2.0` error message will currently be printed out for each file linted. We could just print it once in the beginning, but I think it's fine to leave it as is because we can't lint the file anyway and need to show the user some kind of error.
1. It looks like `babelCore#parseSync` will still parse without a config file, it just won't enable any plugins. Does this seem like a reasonable default? I'm worried about it being confusing behavior since Babel itself won't run any transformations without a config file. An alternative would be to run a check to see if a config file can be found (not sure if `@babel-core` provides an API for this already or not) and throw an error with a helpful message, if not.

Again, I don't think either of these are critical to landing this PR, but I'd love to make sure we decide what we want to do with these cases before we actually make a release. Please feel free to merge this!

kentcdodds:
> An alternative would be to run a check to see if a config file can be found (not sure if @babel-core provides an API for this already or not) and throw an error with a helpful message, if not.

I am in favor of this. If `babel-eslint` cannot find the config, then it's likely someone misconfigured something. Maybe just a warning log would be ok though? And an option to disable the warning could be helpful as well.

kaicataldo:
@kentcdodds How does it sound for me to merge this as-is and then for me to make some issues for the two questions I have above so we can discuss?

existentialism:
@kaicataldo I'm down with that

kentcdodds:
I'm good with that. This would definitely be a nice-to-have :+1:

SimenB:
You can do `loadOptions ` and throw/warn if you don't find any config? If found, pass it as second param to `parse`.

https://babeljs.io/docs/en/babel-core#loadoptions

loganfsmyth:
I agree about filing it separately. The issue is that not all files in your project may have a config. If you have a `.babelrc` in `src/` but didn't bother with any Babel config for the rest of your project, you should still be able to lint those files using the parser's defaults.

kaicataldo:
Great! Will make some issues for us to discuss. Thanks for the input, everyone - it's been super helpful 👍

kentcdodds:
celebretory gif

Woo! Thanks!

kaicataldo:
Issue to discuss how to handle missing config files here: https://github.com/babel/babel-eslint/issues/738

arahansen:
The README changes added here implies this change has already been released (or that is me optimistically reading into the description). Are these changes published anywhere?

kaicataldo:
It hasn't been released yet. We just decided on a path forward for [this last issue](https://github.com/babel/babel-eslint/issues/738) today and, pending landing that, we should be able to get a release out!

hzoo:
Published these changes as `babel-eslint@11.0.0-beta.0` in the next tag for now and we will publish as 11.0.0 if we don't have any other issues

the-spyke:
What if I have no `.babelrc`, but a custom preset? I remember asking about config file (#563), but time flies and approaches evolved since.

the-spyke:
@kaicataldo Often you manually pass options to `babel-loader`. For example as in [create-react-app scripts](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/config/webpack.config.js)

hadnet:
> Published these changes as `babel-eslint@11.0.0-beta.0` in the next tag for now and we will publish as 11.0.0 if we don't have any other issues

When using babel-eslint@11.0.0-beta.0 I got this error -> Parsing error: params is not iterable

kaicataldo:
@hadnet Please create a new issue. Thanks!

hadnet:
@kaicataldo I think somebody already did (#749)

jharris4:
Hi, I'm late to the party here, but we have a setup where we have tools that read our babel config from a custom file which is outside of the babel config file process, kind of akin to a custom preset as mentioned by @the-spyke.

Could we please add support in version `11.x` to allow babelOptions that look something like this:

```
options.babelOptions = {
  presets: [ /* presets go here */ ],
  plugins: [ /* plugins go here */ ]
}
```

If I'm not mistaken, this should be a fairly simple change to make and I'd be willing to attempt a PR.

This would still allow the babel config to be custom specified, but just takes away the requirement that a babel config file has to exist in a specific location/format.

JounQin:
When will it be released as stable???

PR #729, opened 2018-12-05
support classPrivateMethods

neyasov:


SeinopSys:
I'm running into babel/eslint-plugin-babel#166 and now I'm wondering if this would fix it.

hzoo:
Can you include a test case as well to make sure it's fixed? This is correct from a plugin point of view but curious if that's all that's needed.
Something like this

https://github.com/babel/babel-eslint/blob/3f61a68a049293715e4a6bd2549c7fd21c8c14dc/test/babel-eslint.js#L524-L528

and maybe a non-regression test like checking some rules are working

Ateiri:
Hey, not works for me, it produces a lot of weird eslint errors.
But babel compiles private methods without any problems.

cKiril:
Can anyone comment on the status of this? I'm running into this error message with ESLint 5.12

```
ESLint: Parsing error: This experimental syntax requires enabling the parser plugin: 'classPrivateMethods'

  189 |    *                     false otherwise.
  190 |    */
> 191 |   #set(key, value = null) {
      |   ^
  192 |     if (typeof key !== 'string') {
  193 |       return false;
  194 |     }
```

SeinopSys:
@cKiril I tried asking about the same issue [on Stack Overflow](https://stackoverflow.com/q/53675785/1344955) but nobody responded.

kaicataldo:
This should be fixed by https://github.com/babel/babel-eslint/pull/711

hzoo:
From https://github.com/babel/babel-eslint/pull/711, we shouldn't need to update this type of change anymore (babel-eslint should read from your babel config which will turn on or off for babel-eslint which is a better long term solution

WangJiangJiaoZi:
Eslint doesn't work well with Babel and often throws exceptions in vs code.

goldins:
Why not PHP 7.3 or 7.4?

Issue #707, opened 2018-10-28
Remove comment attachment


kaicataldo:
ESLint no longer uses attached comments in the AST and so babel-eslint doing comment attachment is now work that is just thrown away. I'd like to propose removing the behavior entirely (but still keeping the top-level comments array that ESLint _does_ require to calculate comment location internally.

I opened a [similar issue](https://github.com/eslint/espree/issues/394) for the default parser, Espree.

Note that this is a breaking change, but it seems unlikely that any other project will really be affected by it since the purpose of this project is to provide compatibility with ESLint.

PR #736, opened 2019-01-10
Remove comment attachment

kaicataldo:
fixes https://github.com/babel/babel-eslint/issues/707

ESLint no longer uses attached comments provided by the parser. This behavior has already been [removed from the default parser](https://github.com/eslint/espree/pull/405). I also refactored some of this code to hopefully make it clearer/remove some commented out code.


lib/babylon-to-espree/attachComments.js
```diff
@@ -1,59 +0,0 @@
-"use strict";
-
-// comment fixes
-module.exports = function(ast, comments, tokens) {
```
kaicataldo: This was actually now incorrectly named. Rather than doing comment attachment, it's actually calculating the location information for the top-level `Program` node. Renamed this to `convertProgramNode` and encapsulated all the logic of transforming the `Program` node to what ESLint expects in that function.

existentialism: Agreed, much less confusing :)

test/specs/babel-eslint.js
```diff
@@ -50,18 +52,18 @@ function parseAndAssertSame(code) {
     assertImplementsAST(esAST, babylonAST);
   } catch (err) {
     const traversal = err.message.slice(3, err.message.indexOf(":"));
-    err.message += unpad(`
+    err.message += `\n${unpad(`
```
kaicataldo: For ease of reading :)

test/specs/babel-eslint.js
```diff
@@ -13,12 +13,15 @@ function lookup(obj, keypath, backwardsDepth) {
...
 
-  return keypath
-    .split(".")
-    .slice(0, -1 * backwardsDepth)
-    .reduce((base, segment) => {
-      return base && base[segment], obj;
-    });
+  let pathSegments = keypath.split(".");
+
+  if (pathSegments.length > 1) {
```
kaicataldo: There was a bug here where if it's a top level key (i.e. `start`) it would throw and the reported error was a runtime error rather than an assertion error.

test/specs/babel-eslint.js
```diff
@@ -13,12 +13,15 @@ function lookup(obj, keypath, backwardsDepth) {
...
+    pathSegments = pathSegments.slice(0, -1 * backwardsDepth);
+  }
+
+  return pathSegments.reduce((base, segment) => {
+    return base && base[segment], obj;
```
loganfsmyth: Doesn't this end up being `return obj;`, meaning this whole statement is basically

```
return pathSegments.length === 1 ? pathSegments[0] : obj;
```

lib/babylon-to-espree/convertAST.js
```diff
@@ -31,16 +33,15 @@ const astTransformVisitor = {
     node._babelType = node.type;
 
     if (node.innerComments) {
-      node.trailingComments = node.innerComments;
       delete node.innerComments;
     }
 
     if (node.trailingComments) {
-      convertComments(node.trailingComments);
+      delete node.trailingComments;
```
loganfsmyth: Just for my own understanding, does ESLint just not used these properties?

kaicataldo: That's right! We dropped the need for comment attachment to happen at the parser level in v4 (see docs [here](https://github.com/eslint/eslint/blob/caeb223c4f7b0b6fe35e5348ae0df4c6446b5bed/docs/user-guide/migrating-to-4.0.0.md#-ast-nodes-no-longer-have-comment-properties)). We decided to move away from thinking about comments in the context of AST nodes (which had to be reimplemented by every parser and was buggy at best) and instead to just think of them in the context of tokens. ESLint now just takes the tokens and comments arrays provided by the parser and gives rules access to comments through the use of utility methods like `sourceCode#getTokenBefore(nodeOrToken, { includeComments: true })`.

hzoo: I was talking to Kai about being able to turn off even adding the comments in @babel/parser but if the only use case is for babel-eslint we probably don't need to make that option then? Unless there's a case for it being much faster (would need to test)

test/specs/babel-eslint.js
kaicataldo: Yeah, I think so. To be honest, I don't really feel like this `lookup` function provides a whole lot. The error message without modification already reads `Error: At loc.start.line: are different (1 !== 2)`. The extra logic just appends 
```
:
espree:
  'loc'
babel-eslint:
  'loc'
```
to it. Maybe we can just remove that code entirely?

lib/babylon-to-espree/convertAST.js
kaicataldo: I could explore that - I haven't looked at how comment attachment works in @babel/parser in a long time now, but if it's an easy thing to add I'm happy to do it! That said, I don't think it needs to be tied to this release, since it won't result in any user-facing changes.

test/specs/babel-eslint.js
kaicataldo: Yeah, could be great to switch to Jest! I'll make a note of that [here](https://github.com/babel/babel-eslint/issues/739). For now, would you be opposed to just removing this logic from the error message? If so, I can refactor it to the code snippet @loganfsmyth suggested above so we can land it!

existentialism: I'm not opposed to remove!

kaicataldo: Updated

PR #740, opened 2019-01-11
Add code of conduct

kaicataldo:
Noticed this repo was missing a code of conduct. Copied directly from https://github.com/babel/babel/blob/master/CODE_OF_CONDUCT.md.


hzoo:
What do we think about linking to Babel's main one instead?

kaicataldo:
That seems fine to me!

kaicataldo:
Updated - thoughts on the wording?

Issue #738, opened 2019-01-10
Discussion: how do we handle missing Babel config files in v11


kaicataldo:
From https://github.com/babel/babel-eslint/pull/711#issuecomment-453157175

Original question:
> It looks like babelCore#parseSync will still parse without a config file, it just won't enable any plugins. Does this seem like a reasonable default? I'm worried about it being confusing behavior since Babel itself won't run any transformations without a config file. An alternative would be to run a check to see if a config file can be found (not sure if @babel-core provides an API for this already or not) and throw an error with a helpful message, if not.

Sounds like there might issues with any approach that doesn't just parse using defaults, as described [here](https://github.com/babel/babel-eslint/pull/711#issuecomment-453240035) by @loganfsmyth :
> The issue is that not all files in your project may have a config. If you have a .babelrc in src/ but didn't bother with any Babel config for the rest of your project, you should still be able to lint those files using the parser's defaults.

I guess we have to decide if `babel-eslint` is quite literally only for code that is transformed by Babel or if it can also be used for non-transformed code in a project that mostly consists of files that _are_ transformed by Babel. Concrete example I can think of off the top of my head is a tooling config file (`.eslintrc.js`, `babel.config.js`, etc.) that might not have a config.

From a philosophical standpoint, I actually don't think it's unreasonable to say that those files should be parsed with the default parser (which will always be available since it's a dependency of ESLint and is totally possible with ESLint's glob-based configuration). Practically speaking, though, I'm not sure it's a good idea. It increases the overhead for _every_ new project that uses babel-eslint and has any non-transformed JS code.

loganfsmyth:
FWIW despite my previous comment, I'm not against this as a default, as long as there is an opt-out.

We could do:
```
const cfg = babel.loadPartialConfig(/* ... */);
if (!opts.allowNoConfig && cfg && !cfg.hasFilesystemConfig()) {
  throw new Error(`No Babel config file detected for ${cfg.options.filename}. Either disable config file checking with allowNoConfig:false, or configure Babel so that it can find the config files.`)
}
const ast = cfg && babel.parse(cfg.options);
```

kaicataldo:
> Practically speaking, though, I'm not sure it's a good idea. It increases the overhead for every new project that uses babel-eslint and has any non-transformed JS code.

[...]

Issue #739, opened 2019-01-10
How can we improve tests?


kaicataldo:
How can we structure and write tests in a way that allows us to catch integration issues (changes in Babel or ESLint)?

As discussed [here](https://github.com/babel/babel-eslint/issues/62), this repo used to (attempt to) test against ESLint's core rules.

kaicataldo:
As discussed [here](https://github.com/babel/babel-eslint/pull/736/files#r247175369), maybe we can switch from Mocha to Jest?

hzoo:
I'm ok with whatever, if it helps with maintenance to switch just since we are using it in Babel itself that's cool!

More was thinking about some other ideas:

- Is it possible to import tests from both @babel/parser or eslint and run them as integration tests?
- When we add a new syntax to Babel itself, can we make sure we have a test/code to handle it in babel-eslint as well? And even test against the common rules like `unused-vars`. It's easy to be out of sync (this is a general problem with flow/ts/jsx/any syntax in all of Babel though)
- Can we structure the tests such that it should be simple for a new contributor to make a PR to start the work?

PR #743, opened 2019-01-17
Add requireConfigFile option

kaicataldo:
fixes https://github.com/babel/babel-eslint/issues/738

I'm not entirely sure what the best way to go about testing this would be. We don't have tests for a lot of the behavior surrounding config options at the moment, and I'd love to figure out a way to do it. 

As I see it, it seems like a feasible way to test this would be to create fixture dirs for all the various possible configuration variations, run the linter on the fixture dir, and make assertions around that. One thing I'm not entirely sure about is how to test for the absence of a config file (since there's one in `test/` that will be the default if another isn't supplied). I'm hoping there's a way that doesn't feel quite so clunky and repetitive.

Do we want to merge this as is since this stuff hasn't been tested in the past and try to see if we can improve testing in the future with https://github.com/babel/babel-eslint/issues/739? Happy to defer to the team's consensus on this.


lib/analyze-scope.js
```diff
@@ -323,7 +323,7 @@ module.exports = function(ast, parserOptions) {
         parserOptions.ecmaFeatures.globalReturn) === true,
     impliedStrict: false,
     sourceType: ast.sourceType,
-    ecmaVersion: parserOptions.ecmaVersion || 2018,
+    ecmaVersion: parserOptions.ecmaVersion,
```
kaicataldo: This is already set by https://github.com/babel/babel-eslint/pull/743/files#diff-6d186b954a58d5bb740f73d84fe39073R29

README.md
```diff
@@ -53,6 +53,7 @@ With the parser set, your configuration can be configured as described in the [C
 
 Additional configuration options can be set in your ESLint configuration under the `parserOptions` key. Please note that the `ecmaFeatures` config property may still be required for ESLint to work properly with features not in ECMAScript 5 by default.
 
+- `requireConfigFile` (default `true`) can be set to `false` to allow babel-eslint to run on files that do not have a Babel configuration associated with them. This can be useful for linting files that are not transformed by Babel (such as configuration files), though using [glob-based configuration](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) to use the default parser for these files is recommended. Note that when no configuration file is found that babel-eslint will not parse any experimental syntax.
```
existentialism: > Note that when no configuration file is found that babel-eslint will not parse any experimental syntax.

that -> then?

Or maybe better as:

> Note, when no configuration file is found, `babel-eslint` will not parse any experimental syntax.

existentialism: Also, maybe:

> This can be useful for linting files that are not transformed by Babel (such as configuration files), though we recommend using the default parser via [glob-based configuration](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns).

kaicataldo: Updated!

lib/parse.js
```diff
@@ -35,7 +40,24 @@ module.exports = function(code, options) {
...
+      }
+
+      opts = config.options;
+    }
+  }
+
   let ast;
+
   try {
     ast = parse(code, opts);
```
loganfsmyth: Not related to this PR, but do we want to throw a helpful error if this ends up returning `null`? Otherwise I assume it will throw if someone tries to run ESLint on a file that is in the user's `.babelrc`?

README.md
```diff
@@ -53,6 +53,7 @@ With the parser set, your configuration can be configured as described in the [C
 
 Additional configuration options can be set in your ESLint configuration under the `parserOptions` key. Please note that the `ecmaFeatures` config property may still be required for ESLint to work properly with features not in ECMAScript 5 by default.
 
+- `requireConfigFile` (default `true`) can be set to `false` to allow babel-eslint to run on files that do not have a Babel configuration associated with them. This can be useful for linting files that are not transformed by Babel (such as tooling configuration files), though we recommend using the default parser via [glob-based configuration](https://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns). Note: babel-eslint will not parse any experimental syntax when no configuration file is found.
```
loganfsmyth: Should we add some examples of using `overrides` in the ESLint config to target `babel-eslint` for specific files?

kaicataldo: That's a good idea!

kaicataldo:
Thanks! Will follow up with a few smaller PRs to address the suggested additions above.

lib/parse.js
kaicataldo: Question about this: It looks like `babelCore#parseSync` returns `null` when there isn't a config, but since we have a default configuration we pass through if a file isn't found with `requireConfig: false` and it warns when a config isn't found with `requireConfig: true`, I'm not seeing how `babel-eslint` can get into this state.

PR #747, opened 2019-01-28
forward allowAwaitOutsideFunction options to babel

dmail:
**Before this pr**
It seems there is no way to configure `babel-eslint` to allow top level await.

**After this pr**
One could write this inside `.eslintrc.js`
```js
module.exports = {
  parser: "babel-eslint",
  parserOptions: {
    allowAwaitOutsideFunction: true
  }
}
```

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -28,6 +28,7 @@ module.exports = function(code, options) {
     ignore: options.babelOptions.ignore,
     only: options.babelOptions.only,
     parserOpts: {
+      allowAwaitOutsideFunction: options.allowAwaitOutsideFunction,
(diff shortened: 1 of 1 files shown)


kaicataldo:
@dmail What version of `babel-eslint` are you using? `babel-eslint@11.0.0-beta.0` now reads your Babel config file, so I wonder if this is still relevant.

dmail:
I was using `9.0.0` (config file does not seems to be read or configurable in `9.0.0`).
I did not realize `11.0.0-beta.0` was reading babel config file.
So yes, this pr seems useless.

kaicataldo:
Closing this as per the author's comment above. Thanks for the PR, regardless.

Issue #625, opened 2018-06-01
Flow annotations and no-undef + no-unused-vars


pnevyk:
Hi,

I'd like to ask what is the status of handling `no-undef` and `no-unused-vars` of Flow type annotations directly in babel-eslint. If I understand it correctly, you started to use new eslint's scope API in [lib/analyze-scope.js](https://github.com/babel/babel-eslint/blob/master/lib/analyze-scope.js) to implement a custom Referencer which should address these issues (or at least `no-undef`).

With babel-eslint@8.2.3, I identified these snippets as violating `no-undef` rule:

```
opaque type Foo = number;
type Foo = { foo<T>(x: T): T };
declare type Foo = number;
declare interface IFoo {}
```

And these are causing `no-unused-vars` error:

```
declare class Foo {}
declare module Foo {}
declare var Foo: number;
declare function foo(): void;
```

If babel-eslint implements these, corresponding rules in eslint-plugin-flowtype can be deprecated and  removed eventually, as mentioned in [this](https://github.com/gajus/eslint-plugin-flowtype/issues/336) issue.

Is this a possibility or I am missing something?

pascalduez:
I'm having a similar issue:

```js
// @flow

import type { Foo } from './foo';

type Whatever<F = Foo> = {
  foo: F,
};
```

```
error  'Foo' is defined but never used  no-unused-vars
```

kaicataldo:
Thank you for the issue. Now that [`@babel/eslint-parser` has been released](https://babeljs.io/blog/2020/07/13/the-state-of-babel-eslint), we are making this repository read-only. If this is a change you would still like to advocate for, please reopen this in the [babel/babel monorepo](https://github.com/babel/babel).

PR #750, opened 2019-02-10
Fix `no-undef` for several cases

pnevyk:
This PR deals with #625. It fixes three things (implementations are split into separate commits for convenient review):

* opaque type aliases (related PR is #696)
* `declare type = ...` and `declare interface Foo {}` which were missing
* generic parameters for function type annotations

One thing I am not sure about is [this part](https://github.com/babel/babel-eslint/compare/master...pnevyk:missing-type-visitors?expand=1#diff-3ab10361493aeb949966224850773473R299). The root cause of function generics misbehavior was that `_visitTypeAnnotation` was treating type parameters of function type annotation as variable references instead of variable declarations. But I am not sure if this is the best fix, especially whether `_nestTypeParamScope` is used correctly.

I must have change some tests which were imho incorrect ([1](https://github.com/babel/babel-eslint/compare/master...pnevyk:missing-type-visitors?expand=1#diff-4acaf9ae6b90e1b76c40ecff6d815fe0L813), [2](https://github.com/babel/babel-eslint/compare/master...pnevyk:missing-type-visitors?expand=1#diff-4acaf9ae6b90e1b76c40ecff6d815fe0L968)).

---

I would like to tackle the other half of #625 issue as well, either in this PR or in a separate one. But I want to be sure that my reasoning is correct:

In my opinion, `declare ...` statements should not emit `no-unused-vars` regardless if they are referenced in the file or not, since they serve as library definitions (ie. typed interface with outside world for code without Flow types in sources). Therefore I propose to reference these declarations automatically.

---

Also, I think that [this known issue](https://github.com/babel/babel-eslint/blame/35a64b2faecc726d8b7f62b99a5d37c971fe3fa3/README.md#L112) can be removed from the README, as it is not issue anymore for some time already.

### DIFF
--- a/lib/analyze-scope.js
+++ b/lib/analyze-scope.js
@@ -177,6 +177,14 @@ class Referencer extends OriginalReferencer {
     this._visitDeclareX(node);
   }
 
+  DeclareTypeAlias(node) {
+    this._visitDeclareX(node);
+  }
+
+  DeclareInterface(node) {
+    this._visitDeclareX(node);
+  }
+
   // visit OptionalMemberExpression as a MemberExpression.
   OptionalMemberExpression(node) {
     super.MemberExpression(node);
--- a/lib/analyze-scope.js
+++ b/lib/analyze-scope.js
@@ -296,8 +296,16 @@ class Referencer extends OriginalReferencer {
       } else if (propertyType.type === "typeAnnotation") {
         this._visitTypeAnnotation(node.typeAnnotation);
       } else if (propertyType.type === "typeParameters") {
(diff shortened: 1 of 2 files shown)


gajus:
Any reason this has not been merged?

kaicataldo:
Active development [has moved to the `babel/babel` monorepo](https://github.com/babel/babel-eslint#note-babel-eslint-is-being-moved-into-the-babel-monorepo-please-follow-along-here-to-track-our-progress). We are working to get the new packages out the door, and appreciate your patience!

kaicataldo:

PR #753, opened 2019-02-14
Fix to convert hash token.

### LINKED ISSUE #752: The `type` of the hash(#) token is not a string.
I'm using babel-eslint@8.2.6.

Parsing the following source code with `babel-eslint` creates an invalid token.

```js
  export default class TouristNumber extends Vue {
    #pro1='hello'
  }
```

token:

```ts
Token {
  type: 
   TokenType {
     label: '#',
     keyword: undefined,
     beforeExpr: false,
     startsExpr: false,
     rightAssociative: false,
     isLoop: false,
     isAssign: false,
     prefix: false,
     postfix: false,
     binop: null,
     updateContext: null },
  value: undefined,
  start: 92,
  end: 93,
  loc: SourceLocation { start: { line: 4, column: 0 }, end: { line: 4, column: 1 } },
  range: [ 92, 93 ] }
```

The `type` of the token must be a string.
https://eslint.org/docs/developer-guide/working-with-custom-parsers#the-ast-specification

But the result is not a string.

I think that you should add `tt.hash` to [convertToken.js](https://github.com/babel/babel-eslint/blob/master/lib/babylon-to-espree/convertToken.js)

---

ref https://github.com/vuejs/eslint-plugin-vue/issues/814

ota-meshi:
Fixed #752 

This pull request makes the hash token of the class private field to be converted to the token specification of ESLint.

### DIFF
--- a/lib/babylon-to-espree/convertToken.js
+++ b/lib/babylon-to-espree/convertToken.js
@@ -44,6 +44,7 @@ module.exports = function(token, tt, source) {
     type === tt.bang ||
     type === tt.tilde ||
     type === tt.doubleColon ||
+    type === tt.hash ||
     type.isAssign
   ) {
     token.type = "Punctuator";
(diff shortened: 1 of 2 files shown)


hzoo:
looks good, does it cause issues with specific rules with this change? I assume we will have further changes

ota-meshi:
@hzoo Thank you for quickly merging this PR!

Issue #728, opened 2018-12-05
How can I use Private Instance Methods?


weihongyu12:
Babel 7.2.0 released today has officially supported Private Instance Methods, but when I use it, the project can be successfully built, but there is an error message about ESLint:

```
error: Parsing error: This experimental syntax requires enabling the parser plugin: 'classPrivateMethods'
```

The following is my operating environment:

Node.js 10.14.0
npm 6.4.1
Babel 7.2.0
Babel-eslint 10.0.1
Eslint 5.9.0

```JavaScript
// babel.config.js
module.exports = {
  presets: [
    '@vue/app',
  ],
  plugins: [
    '@babel/plugin-proposal-private-methods',
  ],
};
```

```JavaScript
// .eslintrc.js
module.exports = {
  root: true,
  env: {
    node: true,
  },
  extends: [
    'plugin:vue/recommended',
    '@vue/airbnb',
  ],
  rules: {
    'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
    'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
  },
  parserOptions: {
    parser: 'babel-eslint',
  },
};
```

My project is built on Vue, I just want to try Private Instance Methods. I can provide the project code if needed.

cljnilsson:
Been trying to do the same with React, no luck.

weihongyu12:
I think this is not the reason for the framework, #729 may be solving this issue.

existentialism:
This should be fixed by #711

hzoo:
^ fixed by above so should be fixed in the next release (Major)!

Skriptach:
I still got an error
```
Parsing error: D:\Projects\sexynewtab\app\scripts\interface\iconElementComponent.js: Support for the experimental syntax 'classPrivateProperties' isn't currently enabled (6:3):

  4 | 
  5 | 	window.IconElement = class IconElement extends HTMLElement {
> 6 | 		#type;
    | 		^
  7 | 		constructor(type) {
  8 | 			super();
  9 |eslint
```

hadnet:
I added babel private method plugin to am ejected create-react-app project and I get a similar Skriptach error but im my case is about private methods: "Support for the experimental syntax 'classPrivateMethods' isn't currently enabled (10:3):"

hadnet:
[...]

PR #776, opened 2019-06-24
Add failing test for `classPrivateMethods` (regression)

mrtnzlml:
Hi, I read this PR https://github.com/babel/babel-eslint/pull/729 and this issue https://github.com/babel/babel-eslint/issues/728 and it seems like `classPrivateMethods` should work. However, it doesn't. So I added a failing test since I don't know how to fix it.

Thank you very much for checking it.

### DIFF
--- a/test/specs/non-regression.js
+++ b/test/specs/non-regression.js
@@ -1830,6 +1830,14 @@ describe("verify", () => {
     );
   });
 
+  it("works with classPrivateMethods", () => {
+    verifyAndAssertMessages(
+      `
+        class A { #c() {} }
+      `
+    );
+  });
+
   it("works with optionalCatchBinding", () => {
     verifyAndAssertMessages(
       `


anthonator:
Is there a workaround for this until this is fixed?

justintime4tea:
lib/parse.js does not have "classPrivateMethods" listed in the plugins array. It appears that babel-eslint doesn't give a hoot about what you pass it for options either so you can't pass it "classPrivateMethods" to enable it yourself.

The workaround for me (not really practical but makes my editor happy) is to re-add "classPrivateMethods" back into the file... for every project I'm in that it bugs me lol

node_modules/babel-eslint/lib/parse.js
Right around line 21 is an array Plugins
Add it to that.

omerdn1:
This issue is fixed in #801.
Hoping this gets merged soon.

kaicataldo:

Issue #768, opened 2019-05-05
Allow pass options directly in babelOptions rather than a config file


otakustay:
Is it possible to allow babel options in eslint config like:

```js
module.exports = {
    parser: "babel-eslint",
    parserOptions: {
        babelOptions: {
            presets: [
                '@babel/env',
                '@babel/typescript',
            ],
        },
    },
};
```

In this case we can simply wrap configurations in a common module other than placing a standalone babel config file.

This results in a "No Babel config file detected" error currently.

axten:
we strongly need that before version 11.0, we don't have a babelrc file in our projects. Its all configured within webpack babel loader config

axten:
any other thoughts here?

kaicataldo:
Rather than duplicating this configuration, would it be possible to create a separate module for the Babel config that is then imported into the Webpack config and is pointed to in the ESLint config?

The motivation behind the change is to have Babel’s parser have the same configuration during the parsing step for both Babel and ESLint.

axten:
That is exactly what I try to achieve, I now have 3 tools that need a babel config, webpack, jest and eslint. The problem is, that babel in webpack is configured slightly different than in jest. So I already have a custom js file that exports all different kinds of babel config we need in our projects. But to get this approach to work with babel-eslint, I need this feature :)
ATM we're unable to use private class methods proposal because of babel-eslint.

kaicataldo:
Do you mind explaining why giving the path to the config file doesn't work? It seems like what is being proposed here is the opposite of what you described - i.e. having to duplicate your config in the .eslintrc file.

axten:
Thats the key, I don't have a duplicate config ;)
I' have a shared custom config file
```
// babel-configs.js (simple pseudo example)
const common = {}
const webpackModernBrowsers = {}
const webpackLegacyBrowsers = {}
const jest = {}

module.exports = {
    jest: { ...common, ...jest },
   webpackModern: { ...common, ...webpackModernBrowsers },
   webpackLegacy: { ...common, ...webpackLegacyBrowsers },
}
```

nicolo-ribaudo:
I'm curious: why don't you use a normal `babel.config.js` with an `env` block to override, for example, jest-specific plugins?

axten:
Because we use grunt and pass runtime specific options from grunt to the config (debug flag, browserlist)

```
// babel-configs.js (next simple pseudo example)
const common = {}
const webpackModernBrowsers = {}
[...]

Issue #749, opened 2019-02-06
11.0.0-beta.0 - Private Instance Methods - Parsing error: params is not iterable


filmic:
I'm getting following error:
`Parsing error: params is not iterable`

while linting a file with a private instance method.

Using:
`@babel/core@7.2.2`
`@babel/plugin-proposal-class-properties@7.3.0`
`@babel/plugin-proposal-private-methods@7.3.2`
`babel-eslint@11.0.0-beta.0`
`eslint@5.13.0`

`test.js` (linted file):
```
class Person {
  #age = 19;

  #increaseAge() {
    this.#age++;
  }

  birthday() {
    this.#increaseAge();
    alert("Happy Birthday!");
  }
}
```

`.eslintrc.js`:
```
module.exports = {
	"parser": "babel-eslint",
};
```

`babel.config.js`:
```
module.exports = {
	presets: [],
	plugins: [
		"@babel/plugin-proposal-class-properties",
		"@babel/plugin-proposal-private-methods"
	]
};
```

hadnet:
Same thing here.

hadnet:
I resolved that problem adding in .eslintrc `"extends": [
        "plugin:eslint-plugin/classPrivateMethods"
    ]`

and using this versions:
"babel-eslint": "^10.0.1",
"eslint": "^5.13.0"

satazor:
I'm also experiencing the same issue, I got around it temporarily by using arrow functions instead:

```js
 #increaseAge = () => {
    this.#age++;
  }
```

satazor:
Actually, with the arrow function syntax I get:

> Unexpected 'this'. (babel/no-invalid-this)

satazor:
After digging, I found this related issue: https://github.com/babel/babel/issues/9506

filmic:
The issue in Babel (babel/babel#9506) is blocked by estree/estree#180.

IAmJulianAcosta:
I can confirm this:

```
    "@babel/plugin-proposal-class-properties": "^7.3.4",
    "@babel/plugin-proposal-private-methods": "^7.3.4",
    "eslint": "5.15.1",
    "babel-eslint": "11.0.0-beta.0"
```

sepehr:
I also second this:

```
"@babel/plugin-proposal-class-properties": "^7.5.5",
"@babel/plugin-proposal-private-methods": "^7.6.0",
"eslint": "6.1.0",
"babel-eslint": "11.0.0-beta.0"
```

[...]

PR #784, opened 2019-07-26
add support for preset & plugins in babelOptions

jharris4:
As discussed in a comment at the end of #711 I have a use case for the following:

```
parserOptions: {
  requireConfigFile: false,
  babelOptions: {
    presets: [ /* presets go here */ ],
    plugins: [ /* plugins go here */ ]
  }
}
```

The change to the code in this PR was trivial and I confirmed locally that it does work as expected, allowing me to pass in a custom babel config without using a babel config file.

Writing the test case took a couple of hours as I had to figure out a way to show that the change does exactly what is expected.

Feedback on improving/cleaning-up the tests would be more than welcome...

### DIFF
--- a/test/specs/babel-options.js
+++ b/test/specs/babel-options.js
@@ -1,0 +1,161 @@
+"use strict";
+
+const eslint = require("eslint");
+const unpad = require("dedent");
+
+const parser = require("../..");
+
+function verifyAndAssertMessagesWithSpecificESLint(
+  code,
+  rules,
+  expectedMessages,
+  sourceType,
+  overrideConfig,
+  linter
+) {
+  const config = {
+    parser: "current-babel-eslint",
+    rules,
+    env: {
+      node: true,
+      es6: true,
+    },
+    parserOptions: {
+      sourceType,
+      ecmaFeatures: {
+        globalReturn: true,
+      },
+    },
+  };
+
+  if (overrideConfig) {
+    for (const key in overrideConfig) {
+      config[key] = overrideConfig[key];
+    }
+  }
+
+  const messages = linter.verify(code, config);
+
+  if (messages.length !== expectedMessages.length) {
+    throw new Error(
+      `Expected ${expectedMessages.length} message(s), got ${
+        messages.length
+      }\n${JSON.stringify(messages, null, 2)}`
+    );
+  }
+
+  messages.forEach((message, i) => {
+    const formatedMessage = `${message.line}:${message.column} ${
+      message.message
+    }${message.ruleId ? ` ${message.ruleId}` : ""}`;
+    const expectedMessage = expectedMessages[i];
+    if (typeof expectedMessage === "string") {
+      if (formatedMessage !== expectedMessage) {
+        throw new Error(
+          `
+          Message ${i} does not match:
+          Expected: ${expectedMessage}
+          Actual:   ${formatedMessage}
+        `
+        );
+      }
+    } else if (
+      typeof expectedMessage === "object" &&
+      expectedMessage !== null &&
+      typeof expectedMessage.matcher === "function" &&
+      typeof expectedMessage.message === "string"
+    ) {
+      if (!expectedMessage.matcher(formatedMessage)) {
+        throw new Error(
+          `
+          Message ${i} does not match:
+          Expected: ${expectedMessage.message}
+          Actual:   ${formatedMessage}
+        `
+        );
+      }
+    } else {
+      throw new Error(
+        "expected messages must be string or object { matcher: function, message: string }: " +
+          i +
+          " -> " +
+          typeof expectedMessage[i]
+      );
+    }
+  });
+}
+
+function verifyAndAssertMessages(
+  code,
+  rules,
+  expectedMessages,
+  sourceType,
+  overrideConfig
+) {
+  const linter = new eslint.Linter();
+  linter.defineParser("current-babel-eslint", parser);
+
+  verifyAndAssertMessagesWithSpecificESLint(
+    unpad(`${code}`),
+    rules || {},
+    expectedMessages || [],
+    sourceType,
+    overrideConfig,
+    linter
+  );
+}
+
+describe("babel options", () => {
+  it("uses the plugins and presets specified in the options", () => {
+    const code = "class AClass { classProp = 123 };";
+
+    const rules = { "no-unused-vars": 1, semi: 1 };
+
+    const babelOptions = {
+      root: __dirname,
+      rootMode: "root",
+      babelrc: false,
+      presets: ["@babel/preset-env"],
+      plugins: [],
+    };
+
+    verifyAndAssertMessages(
+      code,
+      rules,
+      [
+        {
+          matcher: message => message.startsWith("1:25 Parsing error:"),
+          message: "1:25 Parsing error: <path> <rest of message>",
+        },
+      ],
+      undefined,
+      {
+        parserOptions: {
+          requireConfigFile: false,
+          babelOptions: {
+            ...babelOptions,
+            presets: ["@babel/preset-env"],
+            plugins: [],
+          },
+        },
+      }
+    );
+
+    verifyAndAssertMessages(
+      code,
+      rules,
+      ["1:7 'AClass' is defined but never used. no-unused-vars"],
+      undefined,
+      {
+        parserOptions: {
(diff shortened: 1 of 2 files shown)


jharris4:
@kentcdodds Thanks for reviewing and approving!

Are two approvals needed for a PR to be merged? I don't see anything in [CONTRIBUTING.md](https://github.com/babel/babel-eslint/blob/master/CONTRIBUTING.md).

Would love to be able to use this change once it's merged and released to a new beta... ;-)

kentcdodds:
I'm not really a maintainer. I don't know what the process for things like this are in this repository. So I'll wait for a maintainer to review this.

test/specs/babel-options.js
```diff
@@ -0,0 +1,161 @@
...
+      plugins: [],
+    };
+
+    verifyAndAssertMessages(
+      code,
+      rules,
+      [
+        {
+          matcher: message => message.startsWith("1:25 Parsing error:"),
+          message: "1:25 Parsing error: <path> <rest of message>",
```
kaicataldo: This error message is a bit confusing to me - is that what's coming from Babel parser?

test/specs/babel-options.js
```diff
@@ -0,0 +1,161 @@
...
+        },
+      ],
+      undefined,
+      {
+        parserOptions: {
+          requireConfigFile: false,
```
kaicataldo: Is it expected that this option would always be used in conjunction with `requireConfigFile: false`? I'm curious what the behavior would be when this is `true`.

We should also probably update the README if we're going to make this change.

test/specs/babel-options.js
jharris4: yes, it's the error from Babel.

To show that the new feature is working, I had to devise a test case showing that the babel configuration gets changed by the provided config.

Unfortunately babel's error messages contain the full path to the file (which is machine dependent) so I couldn't just match the strings exactly.

And in fact, showing that the correct lint error is shown in one case (with correct babel config) and that babel errors (and the exact error text doesn't matter) in the other case seemed like an adequate test.

I'm all ears if someone has suggestions for a better way to test the feature though...

test/specs/babel-options.js
jharris4: if you're not using a babel config file then you need to disable `requireConfigFile`. Seemed clear to me, so my thought is that the existing documentation is good enough. ;-)

jharris4:
> One concern I have with this change is that it feels really arbitrary what config options babel-eslint would be allowing and not allowing.

A babel config is a babel config regardless of where it comes from. The only thing changing here is **where** this config gets specified, but the available settings/options are exactly the same.

> I believe both TypeScript and typescript-eslint require a `tsconfig.json`, and this feels like a really similar case to me.

A lot of tools do follow the convention of requiring a configuration file to be named a certain way or for it to be in a specific location. There's nothing wrong with that per se, but it is short sited to force this to be the case (in my opinion) since it limits interoperability with other tools.

Better to support multiple options and let users or build tool authors pick the approach that works best for them... ;-)

axten:
@hzoo Is it possible to merge this and release that as another v11 beta? this way we could test this strongly needed feature... Thanks

jharris4:
@axten, yes it'd be great to get this merged, and then I could tackle making another PR for 11.x to fix the eslint 6.3 issues...

axten:
anything new here? we're still unable to use private class methods because of this not been merged...

nicolo-ribaudo:
@axten Private methods are broken for another reason. If you are already using private methods with Babel, then `babel-eslint` should already be able to parse them.

axten:
we don't have a babel config file, so we're unable to use any newer babel plugin since month.
private class methods are not working for us until #801 is released in the v10.x branch or #768 lands in the v11.x branch.

jharris4:
@axten it's not ideal, but since there hasn't been a new release of this package in quite some time we've been using a fork with this PR included.

It's based off the `11.x` branch, and you can find it on npm as `babel-eslint-fork`.

We've been using it for a few months now without any issues.

axten:
thank you, I will have a look...

hadnet:
> @axten it's not ideal, but since there hasn't been a new release of this package in quite some time we've been using a fork with this PR included.
> 
> It's based off the `11.x` branch, and you can find it on npm as `babel-eslint-fork`.
> 
> We've been using it for a few months now without any issues.

I'm using it but I got the same error: "This experimental syntax requires enabling the parser plugin: 'classPrivateMethods"

jharris4:
@hadnet I haven’t tried classPrivateMethods, but it sounds like you maybe need to tweak your babel config?

If you share your config and a sample of code using private class methods I can take a look...

hadnet:
> @hadnet I haven’t tried classPrivateMethods, but it sounds like you maybe need to tweak your babel config?
> 
> If you share your config and a sample of code using private class methods I can take a look...

```
const developmentEnvironments = ['development', 'test'];

const developmentPlugins = [require('react-hot-loader/babel')];

const productionPlugins = [
  require('babel-plugin-dev-expression'),

  // babel-preset-react-optimize
  require('@babel/plugin-transform-react-constant-elements'),
  require('@babel/plugin-transform-react-inline-elements'),
  require('babel-plugin-transform-react-remove-prop-types')
];

module.exports = api => {
  // see docs about api at https://babeljs.io/docs/en/config-files#apicache

  const development = api.env(developmentEnvironments);

  return {
    presets: [
      [
        require('@babel/preset-env'),
        {
          targets: { electron: require('electron/package.json').version }
        }
      ],
      require('@babel/preset-flow'),
      [require('@babel/preset-react'), { development }]
    ],
    plugins: [
      // Stage 0
      require('@babel/plugin-proposal-function-bind'),

      // Stage 1
      require('@babel/plugin-proposal-export-default-from'),
      require('@babel/plugin-proposal-logical-assignment-operators'),
      [require('@babel/plugin-proposal-optional-chaining'), { loose: false }],
      [
        require('@babel/plugin-proposal-pipeline-operator'),
        { proposal: 'minimal' }
      ],
      [
        require('@babel/plugin-proposal-nullish-coalescing-operator'),
        { loose: false }
      ],
      require('@babel/plugin-proposal-do-expressions'),

      // Stage 2
      [require('@babel/plugin-proposal-decorators'), { legacy: true }],
      require('@babel/plugin-proposal-function-sent'),
      require('@babel/plugin-proposal-export-namespace-from'),
      require('@babel/plugin-proposal-numeric-separator'),
      require('@babel/plugin-proposal-throw-expressions'),

      // Stage 3
      require('@babel/plugin-syntax-dynamic-import'),
      require('@babel/plugin-syntax-import-meta'),
      [require('@babel/plugin-proposal-private-methods'), { loose: true }],
      [require('@babel/plugin-proposal-class-properties'), { loose: true }],
      require('@babel/plugin-proposal-json-strings'),

      ...(development ? developmentPlugins : productionPlugins)
    ]
  };
};
```
When I set my eslintrc to 'parser': 'babel-eslint-fork', I got another error: "Parsing error: params is not iterable."

When I'm using the arrow function way no error at all.

jharris4:
@hadnet I don’t see anything obviously wrong, and I see there’s a related issue: https://github.com/babel/babel-eslint/issues/728

I’m traveling for the holidays but will give it a try when I’m back at my computer.

In the meantime, have you tried switching the order of the `private-methods` and `class-properties` plugins so that class properties is first?

jharris4:
@hadnet I finally had some time to look into using the `@babel/plugin-proposal-private-methods`, and it's installing/activated correctly using this plugin, but I get an error when linting:

```0:0  error  Parsing error: params is not iterable```

It looks like this is a known issue https://github.com/babel/babel-eslint/issues/749, and it seems that it's blocked until a PR gets merged into estree...

axten:
@hzoo @kaicataldo @nicolo-ribaudo this project seems to be unmaintained. I cannot understand why this don't get merged and released. Several people are waiting for this since july...

kaicataldo:
Hey there, apologies for not getting back to you sooner. All active development has moved to the [main Babel monorepo](https://github.com/babel/babel). You can track our progress [here](https://github.com/babel/babel/issues/10752). Do you mind making an issue with your proposal in that repository so we can evaluate it in the context of our v8 release plans?

kaicataldo:
Closing this now that https://github.com/babel/babel/pull/11639 has been released. Thanks for contributing!

PR #785, opened 2019-07-30
Breaking: make AST match Espree 6

### LINKED ISSUE #759: Node type of JSXText is different from espree
For jsx texts like foo in `<>foo</>`, espress has type `JSXText`, while babel-eslint has type `Literal`.

```javascript
require("espree").parse("<>foo</>", { ecmaFeatures: { jsx: true } })
  .body[0].expression.children[0].type; // JSXText


require("babel-eslint").parse("<>foo</>")
  .body[0].expression.children[0].type; // Literal
```

Is this working as intended?

golopot:
This is a breaking change that change the baseline to espree 6.

Some node types has changes in accordance with changes in espree:
- `ExperimentalSpreadProperty` became `SpreadElement`.
- `ExperimentalRestProperty` became `RestElement`.
- `Literal` became `JSXText` (for JSXText).

fixes #759

### DIFF
--- a/lib/babylon-to-espree/convertAST.js
+++ b/lib/babylon-to-espree/convertAST.js
@@ -47,26 +47,6 @@ const astTransformVisitor = {
   exit(path) {
     const node = path.node;
 
-    if (path.isJSXText()) {
-      node.type = "Literal";
-    }
-
-    if (
-      path.isRestElement() &&
-      path.parent &&
-      path.parent.type === "ObjectPattern"
-    ) {
-      node.type = "ExperimentalRestProperty";
-    }
-
-    if (
-      path.isSpreadElement() &&
-      path.parent &&
-      path.parent.type === "ObjectExpression"
-    ) {
-      node.type = "ExperimentalSpreadProperty";
-    }
-
     if (path.isTypeParameter()) {
       node.type = "Identifier";
       node.typeAnnotation = node.bound;
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "eslint-plugin-flowtype": "^3.11.1",
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-prettier": "^3.1.0",
-    "espree": "^3.5.2",
+    "espree": "^6.0.0",
     "husky": "^1.0.0-rc.13",
     "lint-staged": "^7.2.2",
     "mocha": "^6.1.4",
--- a/test/specs/babel-eslint.js
+++ b/test/specs/babel-eslint.js
@@ -451,7 +451,9 @@ describe("babylon-to-espree", () => {
     });
 
     it("super outside method", () => {
-      parseAndAssertSame("function F() { super(); }");
+      assert.throws(() => {
+        parseAndAssertSame("function F() { super(); }");
+      }, /SyntaxError: 'super' keyword outside a method/);
     });
 
     it("StringLiteral", () => {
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -7,6 +7,11 @@
 
 As of the v11.x.x release, babel-eslint now requires Babel as a peer dependency and expects a valid [Babel configuration file](https://babeljs.io/docs/en/configuration) to exist. This ensures that the same Babel configuration is used during both linting and compilation.
 
+Some node types has changes in accordance with changes in espree:
```
kaicataldo: I don’t think we actually need to put this in the readme since it’s only relevant to people upgrading to v11. (though I think it should be documented in a change log or release notes when v11 is finalized).

lib/babylon-to-espree/convertAST.js
```diff
@@ -47,26 +47,6 @@ const astTransformVisitor = {
...
-    if (path.isJSXText()) {
-      node.type = "Literal";
-    }
-
-    if (
-      path.isRestElement() &&
-      path.parent &&
-      path.parent.type === "ObjectPattern"
-    ) {
-      node.type = "ExperimentalRestProperty";
```
kaicataldo: Is`Array{Pattern|Expression}` already taken care of?

README.md
golopot: Updated.

lib/babylon-to-espree/convertAST.js
golopot: Yes.

https://github.com/babel/babel-eslint/blob/2540427c555cf707ced5b71b1717ac401fbcf3d0/test/specs/babel-eslint.js#L485-L495

golopot:
> Can we mark this as a breaking change rather than a feature?

How about this commit message title?
Breaking: update to espree@6

> Also, can we add some tests for these changes?

The old tests has covered the changes. The function `parseAndAssertSame` compare the ast generated by babel and espree.

https://github.com/babel/babel-eslint/blob/2540427c555cf707ced5b71b1717ac401fbcf3d0/test/specs/babel-eslint.js#L485-L495

lib/babylon-to-espree/convertAST.js
kaicataldo: Great, thanks - makes sense!

kentcdodds:
I'm not sure how we're managing the alpha release. Should this be merged into master, or into some interim branch?

kaicataldo:
I'm not sure either - hopefully someone from the core Babel team can help us out here!

golopot:
v11 works are already present in master, so I suppose v11 development happens in master. And v10 development should be done in branch [10.x](https://github.com/babel/babel-eslint/tree/10.x).

kaicataldo:
Friendly ping @nicolo-ribaudo (just because you gave a thumbs up above 😄). This seems like a good thing to land in v11 to me, but would love input from the core team!

kaicataldo:
One more friendly ping here. Our last release ended up causing one of the core rules to crash with `babel-eslint` because I mistakenly removed a check for `ExperimentalRestElement` nodes 😬. Aligning the AST here would help prevent those kinds of bugs going forward!

kaicataldo:
Friendly ping - is this something we can merge?

existentialism:
Do we want to ship 11 and this as 12? Or just lump it in?

kaicataldo:
I think it's fine to lump this in with 11 since it's still in beta.

existentialism:
@kaicataldo let's do it

Issue #791, opened 2019-08-18
ESLint 6.2.0 + babel-eslint + no-unused-vars false positive with for-in loop


feross:
There's a new issue in ESLint 6.2.0 just caught by the `standard` test suite. This may be an issue with `babel-eslint`, though I am not sure since the issue did not exist in ESLint 6.1.0. Here's a pointer to the issue I filed with ESLint: https://github.com/eslint/eslint/issues/12117

yyfearth:
Seems the `eslint-scope` is outdated for the latest eslint.

jharris4:
FYI I tried upgrading `eslint-scope` to latest on my fork of this package, and the false positives are still happening.

DullReferenceException:
Maybe related, having the same problem with `for..of` loops:

```js
import scenarios from './scenarios';

for (const scenario of scenarios) {
  doSomethingWithScenario(scenario);
}
```

Get an eslint error:

> 'scenario' is defined but never used     eslint(no-unused-vars)

yyfearth:
According to the original eslint issue eslint/eslint#12117

> Most likely it is that commit, with babel-eslint in the keyword example there is a `TDZ` scope between function and for, probably because it's using the old version of eslint-scope

feross:
Just wanted to note that ESLint 6.2.1 contains an important security fix (see: https://eslint.org/blog/2019/08/eslint-v6.2.1-released) but it's not possible to upgrade to 6.2.0 or later until this issue is fixed.

JLHwung:
For those who are curious about why ESLint 6.2 + babel-eslint does not work and how we may solve it on `babel-eslint` side, I have written a bit on https://github.com/babel/babel-eslint/pull/793#issuecomment-524435971.

SalvatorePreviti:
Quick question/proposal: `eslint-scope` is already a dependency of `eslint`, would not make sense to use the one shipped with `eslint`, moving `eslint-scope` from `dependencies` to `peer-dependencies` of `babel-eslint`?

hzoo:
Ok merged https://github.com/babel/babel-eslint/pull/794 thanks to everyone and @JLHwung for the PR, released as 10.0.3. Not an easy package to maintain for sure given it's patching the packages themselves. Ideally we'll have better hooks for these things in the future.

Can also do one for v11 beta.

I verifed with one of the test cases on 10.0.2 and it failing and passing with 10.0.3. Let us know if this issue still persists! Thanks!

[...]

PR #792, opened 2019-08-19
Fix ESLint 6.2.0 + babel-eslint + no-unused-vars false positive with for-in loop

przemyslawzalewski:
In order to reach Espree 6 compatibility, I have manually included the changes by proposed by [golopot](https://github.com/golopot) at https://github.com/babel/babel-eslint/pull/785. This is a candidate for a major release.
I have fixed a single broken test case of a super call being called from regular function and added two test cases - one that validates the `no-unused-vars` works as expected when there is a real issue and another for the false-positive case that caused the breakage.
To make the linting and tests passes on the repo itself, I have removed [self-reference on a broken `babel-eslint` version](https://en.wikipedia.org/wiki/Chicken_or_the_egg) by just extending the [base config](https://github.com/babel/eslint-config-babel) with the current one.
These changes fix the false positives introduced with ESLint 6.2.0 and they occur no more when using the forked version.

Fixes:
https://github.com/babel/babel-eslint/issues/791


karolyi:
@hzoo @sebmck @danez let's merge this ASAP, shall we.

nick:
The corresponding [issue](https://github.com/eslint/eslint/issues/12117) on eslint has 94 upvotes so seem this issue affects a lot of users. Any chance we can get this merged soon? ☺️

tmorehouse:
Any news on when this can be merged and released?

przemyslawzalewski:
I have proposed [another PR as a PATCH change](https://github.com/babel/babel-eslint/pull/793) that should be easier to integrate and I hope it will be merged soon.
In the meanwhile, I have released a forked package for the ^10 version with the fixes included:

https://www.npmjs.com/package/@przemyslawzalewski/babel-eslint/v/10.0.3

Usage is simple as changing some entires in a package.json:
`"babel-eslint": "^10.0.2"` -> `"@przemyslawzalewski/babel-eslint": "^10.0.3"` and 
`"parser": "babel-eslint"` -> `"parser": "@przemyslawzalewski/babel-eslint"` within the eslint config file.

There is also a fork of the not yet released ^11 branch, however, it may include some breaking changes:
https://www.npmjs.com/package/@przemyslawzalewski/babel-eslint/v/11.0.0

hzoo:
Thanks everyone for the help and @przemyslawzalewski for the PRs! Went ahead and ended up going with @JLHwung's PR https://github.com/babel/babel-eslint/pull/794 which uses ESLint's deps instead of going with peerDeps since it really depends on the version being used and we don't want users to have to install it directly on their own.

Comment here: https://github.com/babel/babel-eslint/issues/791#issuecomment-524664585

PR #794, opened 2019-08-24
fix: require eslint dependencies from eslint base (v10 backport)

### LINKED ISSUE #791: ESLint 6.2.0 + babel-eslint + no-unused-vars false positive with for-in loop

JLHwung:
Fixes #791 

This PR is inspired from the practice and discussion of #793. Thank you @nicolo-ribaudo for the idea and thank you everyone in the community from which I have learned a lot.

### DIFF
--- a/lib/analyze-scope.js
+++ b/lib/analyze-scope.js
@@ -1,10 +1,14 @@
 "use strict";
 
 const t = require("@babel/types");
-const escope = require("eslint-scope");
-const Definition = require("eslint-scope/lib/definition").Definition;
-const OriginalPatternVisitor = require("eslint-scope/lib/pattern-visitor");
-const OriginalReferencer = require("eslint-scope/lib/referencer");
+const requireFromESLint = require("./require-from-eslint");
+
+const escope = requireFromESLint("eslint-scope");
+const Definition = requireFromESLint("eslint-scope/lib/definition").Definition;
+const OriginalPatternVisitor = requireFromESLint(
+  "eslint-scope/lib/pattern-visitor"
+);
+const OriginalReferencer = requireFromESLint("eslint-scope/lib/referencer");
 const fallback = require("eslint-visitor-keys").getKeys;
 const childVisitorKeys = require("./visitor-keys");
 
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "@babel/parser": "^7.0.0",
     "@babel/traverse": "^7.0.0",
     "@babel/types": "^7.0.0",
-    "eslint-scope": "3.7.1",
-    "eslint-visitor-keys": "^1.0.0"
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -15,8 +15,8 @@
     "@babel/parser": "^7.0.0",
     "@babel/traverse": "^7.0.0",
     "@babel/types": "^7.0.0",
-    "eslint-scope": "3.7.1",
-    "eslint-visitor-keys": "^1.0.0"
+    "eslint-visitor-keys": "^1.0.0",
```
JLHwung: Ideally we should not pin `eslint-visitor-keys` either. But `eslint` does not import `eslint-visitory-keys` until 4.14.0. As we have specified eslint peerDependencies to `>=4.12.1`, I would rather leave it here and then we can remove it on v11.

hzoo:
Awesome, thanks for the work @JLHwung! Like he said, for everyone's reports/discussion earlier!

lib/analyze-scope.js
```diff
@@ -1,10 +1,14 @@
...
-const OriginalPatternVisitor = require("eslint-scope/lib/pattern-visitor");
-const OriginalReferencer = require("eslint-scope/lib/referencer");
+const requireFromESLint = require("./require-from-eslint");
+
+const escope = requireFromESLint("eslint-scope");
+const Definition = requireFromESLint("eslint-scope/lib/definition").Definition;
+const OriginalPatternVisitor = requireFromESLint(
+  "eslint-scope/lib/pattern-visitor"
+);
+const OriginalReferencer = requireFromESLint("eslint-scope/lib/referencer");
```
mysticatea: Personally, I don't think that to depend on the internal files of ESLint is a good idea because it's not a public API. ESLint can change the internal structure even in patch releases (ESLint can major update espree/eslint-scope in patch releases if it doesn't affect public behaviors). Instead, it's better if `babel-eslint` depends on own `espree`/`eslint-scope`. Those packages follow semantic versioning.

In my 2 cents.

mysticatea: As just a fact, to depend on internal files of ESLint is really fragile. It's possible to be broken on every ESLint release, including patch releases.

As my opinion, it's not good that `babel-eslint` adopts such a fragile implementation. I agree with @kaicataldo. I recommend that `babel-eslint` follows the semantic versioning and update the supported ESLint versions in major versions if needed. The latest `eslint-scope` should be compatible with `eslint@>=5.0.0`.

PR #807, opened 2019-11-08
Override plugin if available in eslint config

rajasekarm:
Fixes - https://github.com/babel/babel/issues/10675


kaicataldo:

PR #800, opened 2019-09-15
Add into 11.0.0 the fix from 10.0.2 to 10.0.3

agalatan:
**This PR just patches 11.0.0-beta.0 and versions it to 11.0.0-beta.1, in the exact same way 10.0.2 was patched and became 10.0.3**

*Problem*:
- _babel-eslint@10.0.3_ is being super weird for me (both _eslint 5 & 6_), in some Flow "edge cases",  by reporting **no-undef**. But miraculously, _babel-eslint@11.0.0-beta.0_ fixes that problem. 
- Unfortunately, _babel-eslint@11.0.0-beta.0_ exhibits (for _eslint < 6.2.0_) the exact same issue as _babel-eslint@10.0.2_ did, which was fixed here:  [fix from 10.0.2 => 10.0.3](https://github.com/babel/babel-eslint/commit/354953da5347a005fe566cc7dd338974aab50908)

*Solution*:
- Looking in the code for _babel-eslint@11.0.0-beta.0_, one can see [this code is missing](https://github.com/babel/babel-eslint/commit/354953da5347a005fe566cc7dd338974aab50908).
 
- All I did in this PR was **copy-paste**, **yarn**, and **version it to 11.0.0-beta.1**

PS: This is my first PR

PR #810, opened 2019-12-03
chore: bump to eslint-scope 5 and eslint 6

JLHwung:
This PR is meant to be a hot fix when we are in the transition from `babel-eslint` to `@babel/eslint-parser`. Dependencies are bumped to their latest versions so that it can work with `no-unused-rules` in eslint 6.

Closes #795 , closes #800

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "babel-eslint",
-  "version": "11.0.0-beta.0",
+  "version": "11.0.0-beta.1",
   "description": "Custom parser for ESLint",
   "author": "Sebastian McKenzie <sebmck@gmail.com>",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -22,19 +22,19 @@
     "changelog": "git log `git describe --tags --abbrev=0`..HEAD --pretty=format:' * %s (%an)' | grep -v 'Merge pull request'"
   },
   "engines": {
-    "node": ">=6"
+    "node": ">=8"
```
JLHwung: `node` is bumped due to `eslint-scope 5.0.0` requires `>=8`.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "babel-eslint",
-  "version": "11.0.0-beta.0",
+  "version": "11.0.0-beta.1",
```
JLHwung: I manually bumped this version because
1) 11.0.0-beta.1 has been published from babel repo
2) the old publish process in this repo still works if it starts from `beta.1`

JLHwung:
@kaicataldo Yeah, we have included this PR in https://github.com/babel/babel/pull/10753/files#diff-636618e733ad5d381a4458580edc65b2

PR #812, opened 2019-12-06
Update Babel to ^7.7.0 and enable Flow enums parsing

gkz:
The ability to parse Flow enums was added in Babel 7.7.0. It sits behind an option "enums" for the flow Babel plugin.

Update the Babel dependency to ^7.7.0, fix a small test issue, then enable `enums: true`.

I understand that we wouldn't need to do this with the 11.x series of babel-eslint because Babel is a peer dependency in 11.x, but that's still in beta and not officially released, so that's why I'm opening up a PR for the 10.x series.

### DIFF
--- a/test/babel-eslint.js
+++ b/test/babel-eslint.js
@@ -261,11 +261,11 @@ describe("babylon-to-espree", () => {
   });
 
   it("export named", () => {
-    parseAndAssertSame("export { foo };");
+    parseAndAssertSame("const foo = 1; export { foo };");
   });
 
   it("export named alias", () => {
-    parseAndAssertSame("export { foo as bar };");
+    parseAndAssertSame("const foo = 1; export { foo as bar };");
   });
 
   // Espree doesn't support the optional chaining operator yet
@@ -296,6 +296,16 @@ describe("babylon-to-espree", () => {
       eslintScopeManager: true,
     }).ast;
     assert.strictEqual(babylonAST.tokens[1].type, "Punctuator");
+  });
+
+  // Espree doesn't support Flow enums yet
+  it("flow enums", () => {
+    const code = "enum E {A, B}";
+    const babylonAST = babelEslint.parseForESLint(code, {
+      eslintVisitorKeys: true,
+      eslintScopeManager: true,
+    }).ast;
+    assert.strictEqual(babylonAST.body[0].type, "EnumDeclaration");
   });
 
   it.skip("empty program with line comment", () => {
--- a/test/babel-eslint.js
+++ b/test/babel-eslint.js
@@ -298,7 +298,7 @@ describe("babylon-to-espree", () => {
     assert.strictEqual(babylonAST.tokens[1].type, "Punctuator");
   });
 
-  // Espree doesn't support Flow enums yet
+  // Espree doesn't support Flow enums
   it("flow enums", () => {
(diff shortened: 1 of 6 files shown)


test/babel-eslint.js
```diff
@@ -298,6 +298,16 @@ describe("babylon-to-espree", () => {
     assert.strictEqual(babylonAST.tokens[1].type, "Punctuator");
   });
 
+  // Espree doesn't support Flow enums yet
```
kaicataldo: It's unlikely Espree will ever support Flow (there are no plans to do so at the moment, at least).

```suggestion
  // Espree doesn't support Flow enums
```

gkz:
The support for parsing enums should not be breaking, since `enum` is a "future reserved word" and its usage was previously a parse error.

gkz:
Hi @kaicataldo, now that the holiday season has passed, I was wondering if you'd be willing to take another look at this.

kaicataldo:
@gkz Would you be willing to make this PR over in https://github.com/babel/babel/tree/master/eslint/babel-eslint-parser?  I'm working to get `babel-eslint` back to a sane state for the Babel v8 release, and doing all active development there now. Thanks!

kaicataldo:
Though, actually, this should already work in the latest version. Adding these tests could be helpful though.

gkz:
We need to add the "flow" plugin to the parser options (to be able to parse Flow code), and add the `enums: true` flow plugin option to parse neums. 

I'll take a look

kaicataldo:
@gkz The latest version reads your Babel config at parse time, so if Babel is able to parse it, ESLint should also.

gkz:
@kaicataldo We are blocked on launching the enums feature because of lack of Eslint support. `@babel/eslint-parser` has not been released yet. I would appreciate it if you would consider merging this and releasing 10.0.4. Since the version of `@babel/parser` is specified as `^7.0.0`, anyone installing `babel-eslint` will get the latest version of Babel parser already, so the only change in this PR is adding `{enums: true}`. Since `enum` is a reserved word, any usage would be a parse error.

gkz:
I realized I need to add EnumDeclarations to the scope analysis - will update the PR soon

test/non-regression.js
```diff
@@ -229,6 +229,19 @@ describe("verify", () => {
...
+    it("enum declaration", () => {
+      verifyAndAssertMessages(
+        `
+          enum E {
+            A,
+            B,
+          }
+          E.A;
+        `,
+        { "no-unused-vars": 1, "no-undef": 1 }
```
nicolo-ribaudo: Can you also add a test for an unused enum, which should be reported?

gkz: OK, will add!

gkz:
I have updated the test

kaicataldo:
Do we also need to update the peer dependency on `@babel/core`?

nicolo-ribaudo:
No, it would be a breaking change. If someone is using an old version of Babel, this PR is a noop.

kaicataldo:
Ah right, this is in v10. Got it :+1:

kaicataldo:
Thanks for contributing!

gkz:
Thanks for mergin!

kaicataldo:
Published in `babel-eslint@10.1.0`.

PR #830, opened 2020-04-15
Support custom parserOpts for babel parser

vikr01:
Not sure if I should PR this here or to the monorepo?

This supports custom `parserOpts` being passed in to the babel parser through the eslint config.


kaicataldo:

